PR #5, opened 2016-04-21
Add 0xb demo bits

lukewagner:
This PR adds a (currently candidate) .wasm binary that matches the 0xb branch binary encoding in [design](https://github.com/WebAssembly/design).  The only thing missing (which I'll upload tomorrow assuming [design/#664](https://github.com/WebAssembly/design/pull/664) is merged) is the `br_table` value.  I confirmed locally that it runs in the three configurations: no wasm (which just has the message), FF nightly (0xa), FF with [patches](https://bugzilla.mozilla.org/show_bug.cgi?id=1263202) (0xb).

Happy to get any feedback and update the PR with new binary bits if any issues are found.  I'll of course wait to confirm this decodes in at least one other engine before merging.  (cc @binji @kripken @sunfishcode )

### DIFF
--- a/demo/AngryBots/index.html
+++ b/demo/AngryBots/index.html
@@ -101,8 +101,12 @@ window.addEventListener('resize', softFullscreenResizeWebGLRenderTarget);
     progress: (new UnityProgress(canvas)),
   };
 
+  var version = "";
+  if (Wasm && Wasm.experimentalVersion)
+    version = String(Wasm.experimentalVersion);
+
   var xhr = new XMLHttpRequest();
-  xhr.open('GET', 'AngryBots.wasm', true);
+  xhr.open('GET', 'AngryBots' + version + '.wasm', true);
   xhr.responseType = 'arraybuffer';
   xhr.onload = function() {
     Module.wasmBinary = xhr.response;
--- a/demo/AngryBots/index.html
+++ b/demo/AngryBots/index.html
@@ -102,8 +102,9 @@ window.addEventListener('resize', softFullscreenResizeWebGLRenderTarget);
   };
 
   var version = "";
-  if (Wasm && Wasm.experimentalVersion)
+  if (Wasm && Wasm.experimentalVersion) {
     version = String(Wasm.experimentalVersion);
+  }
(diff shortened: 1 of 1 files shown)


demo/AngryBots/index.html
```diff
@@ -101,8 +101,12 @@
     progress: (new UnityProgress(canvas)),
   };
 
+  var version = "";
+  if (Wasm && Wasm.experimentalVersion)
```
kripken: lgtm. feel free to ignore, but i would be happier with `{}` here

lukewagner: sure; the style in this file is all over

lukewagner:
Now with `br_table` values.  @kripken says SM bits run in Binaryen's 0xb interpreter.  @binji / @titzer: lgty?

titzer:
I haven't tried the demo bits yet but we are basically there functionally.
I'll take a look at the bits tomorrow.

titzer:
Have you had a look at the burndown list recently?

lukewagner:
@titzer Yes, it looks like the only diff from what is "titzer:CL" is the new `form` field in the types section.  Once you add that, I expect V8 should decode these bits.

titzer:
Looks like we never landed a design PR for arities in call, return, and break?

titzer:
We need to rebase #595 onto binary_0xb and land that.

lukewagner:
Alright, assuming @rossberg-chromium can get that rebased, here are the updated bits with arities (and, to be clear, the `nop` values removed from branches/return when arity=0).

lukewagner:
How's it looking?

s3ththompson:
@titzer It looks like we still need to land the `experimentalVersion` property?

titzer:
LGTM.

I've verified that Chrome is able to run these demo bits with this patch (
https://codereview.chromium.org/1896863003/).

-B

titzer:
I also verified the backup path works (i.e. falling back to 0xa) in Chrome.

lukewagner:
Commence landing!

PR #7, opened 2016-06-09
AB wast with explicit drop

kripken:
This updates the AB wast.gz with a version with explicit drop and tee etc. This was done with a hackish binaryen pass that loads the old format and emits the new (in `drop` branch).

Note: not fully tested, wast might be too big for my 32-bit machine here. I did verify smaller things like hello world up to libc++ pass, so it should be ok, unless it hits some corner case.


kripken:
Added optimized wast as well (unoptimized doesn't use tee, as that depends on opts).

s3ththompson:
@kripken is this something you still want to merge?

kripken:
I made it by request from @lukewagner I believe, for testing purposes. I think at this point we need a new version anyhow due to spec changes. @sunfishcode might have such a binary.

PR #8, opened 2016-08-08
make IRC link clickable

marianoguerra:
it takes you to kiwi IRC web client to make it easy for non IRC users to connect and interact.


jfbastien:
Seems useful, thanks!

PR #9, opened 2016-08-17
Add a page with step-by-step instructions to get the WebAssembly tool…

RebeccaRGBA:
…chain running and compiling.

### DIFF
--- a/stepbystep/index.html
+++ b/stepbystep/index.html
@@ -75,14 +75,16 @@ $ PATH=`pwd`/cmake-3.6.0-Darwin-x86_64/CMake.app/Contents/bin:$PATH
           There is one change we need to make, however: when calling cmake, we need to add
           the following flag: <code>-DLLVM_EXPERIMENTAL_TARGETS_TO_BUILD=WebAssembly</code>
         </p>
+        <p>
+          (Also, while the getting started page uses the LLVM svn repositories,
+          here we will use the <a href="https://github.com/llvm-mirror">GitHub LLVM mirror</a>.)
+        </p>
         <pre>
-$ svn co http://llvm.org/svn/llvm-project/llvm/trunk llvm
+$ git clone https://github.com/llvm-mirror/llvm.git llvm
 $ cd llvm/tools
-$ svn co http://llvm.org/svn/llvm-project/cfe/trunk clang
-$ cd clang/tools
-$ svn co http://llvm.org/svn/llvm-project/clang-tools-extra/trunk extra
-$ cd ../../../projects
-$ svn co http://llvm.org/svn/llvm-project/compiler-rt/trunk compiler-rt
+$ git clone https://github.com/llvm-mirror/clang.git clang
+$ cd ../projects
+$ git clone https://github.com/llvm-mirror/compiler-rt.git compiler-rt
 $ cd ../..
 $ mkdir llvm-build
 $ cd llvm-build
--- a/stepbystep/index.html
+++ b/stepbystep/index.html
@@ -93,7 +93,7 @@ $ cd ..
         <h3>Install the correct version of emscripten</h3>
         <p>
           To compile to WebAssembly, we need the incoming branch of emscripten. We can install this through the
-          <a href="http://kripken.github.io/emscripten-site/docs/getting_started/downloads.html">emscripten SDK</a>;
+          <a href="https://kripken.github.io/emscripten-site/docs/getting_started/downloads.html">emscripten SDK</a>;
           however, we need to switch the branch the SDK downloads and installs from. We also need to change
           emscripten's configuration to use the version of LLVM we compiled in the previous section.
         </p>
--- a/stepbystep/index.html
+++ b/stepbystep/index.html
@@ -29,7 +29,11 @@
         <p>
           This page aims to provide complete, working step-by-step instructions
           to compile a simple program directly to WebAssembly.
+        </p>
         <p>
+          The instructions on this page are applicable to Linux and Mac OS X
+          systems. Similar instructions for Windows systems are forthcoming.
+        </p>
         <h3>Install the correct version of cmake</h3>
         <p>
           To compile the rest of the tools, cmake is required. Specifically,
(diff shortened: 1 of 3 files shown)


stepbystep/index.html
```diff
@@ -0,0 +1,147 @@
...
+        <pre>
+          $ curl https://cmake.org/files/v3.6/cmake-3.6.0-Linux-x86_64.tar.gz \
+                 -o cmake-3.6.0-Linux-x86_64.tar.gz
+          $ gunzip cmake-3.6.0-Linux-x86_64.tar.gz
+          $ tar -xf cmake-3.6.0-Linux-x86_64.tar
+          $ PATH=`pwd`/cmake-3.6.0-Linux-x86_64/bin:$PATH
+        </pre>
+        <h3>Install the correct version of LLVM</h3>
+        <p>
+          To compile to WebAssembly, emscripten requires a "vanilla" build of LLVM
```
kripken: A vanilla build is one option of 2, the other is to use emscripten's LLVM and asm2wasm. That's currently the [recommended route](https://github.com/kripken/emscripten/wiki/WebAssembly) as the wasm backend is not stable yet.

I think changing to that would be a simplification here - there is no need to get and build LLVM by itself (the emsdk will do that), and there is no need for the `EMCC_WASM_BACKEND` env var.

jfbastien:
I'm not sure the main wasm site is the right place for this info. I'd rather see it on binaryen / emscripten, and link to it from here. It'll less likely to bitrot.

s3ththompson:
([Live preview](https://rawgit.com/RebeccaRGBA/webassembly.github.io/86f9c08af7830a8f7e8b4f422859d0805376c81b/stepbystep/index.html) with [rawgit.com](rawgit.com) for review convenience)

Looks great. I added a bit of styling around pre/code blocks to the site custom css file (#10). Once it gets merged you can rebase on top of it and pick up the spacing/color fixes.

@jfbastien I actually think the wasm site should always have baseline instructions to get started so that new users don't have to dig around in the docs of a bunch of other tool sites. Once other toolchains emerge we can highlight quick instructions for them too. Detailed instructions can always be housed on other sites, but I think we can handle maintenance for this level of detail.

@RebeccaRGBA did you get a chance to test these instructions on OSX? We might want to make the instructions more general for at least Unix architectures (and Windows in the future).

stepbystep/index.html
```diff
@@ -0,0 +1,147 @@
...
+          the following flag: <code>-DLLVM_EXPERIMENTAL_TARGETS_TO_BUILD=WebAssembly</code>
+        </p>
+        <pre>
+          $ svn co http://llvm.org/svn/llvm-project/llvm/trunk llvm
+          $ cd llvm/tools
+          $ svn co http://llvm.org/svn/llvm-project/cfe/trunk clang
+          $ cd clang/tools
+          $ svn co http://llvm.org/svn/llvm-project/clang-tools-extra/trunk extra
+          $ cd ../../../projects
+          $ svn co http://llvm.org/svn/llvm-project/compiler-rt/trunk compiler-rt
```
jfbastien: These should probably point at the git mirrors instead, since they're https: https://github.com/llvm-mirror

stepbystep/index.html
```diff
@@ -0,0 +1,147 @@
...
+          $ mkdir llvm-build
+          $ cd llvm-build
+          $ cmake -G "Unix Makefiles" -DLLVM_EXPERIMENTAL_TARGETS_TO_BUILD=WebAssembly ../llvm
+          $ make
+          $ cd ..
+        </pre>
+        <h3>Install the correct version of emscripten</h3>
+        <p>
+          To compile to WebAssembly, we need the incoming branch of emscripten. We can install this through the
+          <a href="http://kripken.github.io/emscripten-site/docs/getting_started/downloads.html">emscripten SDK</a>;
```
jfbastien: https

jfbastien:
@s3ththompson if you sign up to keep it up to date then I'm happy ;-)

stepbystep/index.html
RebeccaRGBA: That might be confusing, though, since the guide we just linked to uses the svn repositories.

stepbystep/index.html
RebeccaRGBA: Done.

RebeccaRGBA:
I made the changes I discussed with @s3ththompson and I am going through this process on OS X as well so I can make sure it works there.

stepbystep/index.html
jfbastien: I would rather have the confusing but secure one: github can be obtained over https but LLVM's svn cannot.

stepbystep/index.html
```diff
@@ -0,0 +1,166 @@
...
+          compile time.
+        </p>
+        <p>
+          The <a href="http://clang.llvm.org/get_started.html">clang getting started page</a>
+          provides an excellent step-by-step guide to checking out and compiling LLVM with clang.
+          There is one change we need to make, however: when calling cmake, we need to add
+          the following flag: <code>-DLLVM_EXPERIMENTAL_TARGETS_TO_BUILD=WebAssembly</code>
+        </p>
+        <pre>
+$ svn co http://llvm.org/svn/llvm-project/llvm/trunk llvm
```
jfbastien: Hmm, comment got hidden by github because of update. Reposting:

I would rather have the confusing but secure repo: github can be obtained over https but LLVM's svn cannot.

RebeccaRGBA: Done. We don't have a git mirror of clang-tools-extra but I suppose we don't need that?

jfbastien: Correct, I don't think we do.

jfbastien:
lgtm, ty!

kripken:
I commented [earlier](https://github.com/WebAssembly/webassembly.github.io/pull/9#discussion_r75223987), maybe it was missed, that I don't think we should recommend `EMCC_WASM_BACKEND=1` or building vanilla LLVM, as that path is not stable yet.

Instead, we can just remove those parts from the text (as the default path is the stable and recommended `asm2wasm` one). It also simplifies and shortens the text as well.

s3ththompson:
@kripken Oops! Can you submit a PR?

kripken:
Sure: #11.

PR #11, opened 2016-08-25
Recommend the current stable compiler path

kripken:
It's better to not recommend the wasm backend yet, as it isn't stable. It is also more steps to set up.

By default `emcc` uses `asm2wasm` which is stable.

### DIFF
--- a/stepbystep/index.html
+++ b/stepbystep/index.html
@@ -34,70 +34,11 @@
           The instructions on this page are applicable to Linux and Mac OS X
           systems. Similar instructions for Windows systems are forthcoming.
         </p>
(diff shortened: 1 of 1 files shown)


kripken:
Ping? @s3ththompson 

This seems like a no-brainer. Recommending the wasm backend right now means people are likely to run into bugs and limitations, which would reflect poorly on WebAssembly. Instead, we have a stable option for them to use.

s3ththompson:
LGTM, I chatted offline with @RebeccaRGBA about the changes

I'm going to start a second PR with a toggle for the upstream LLVM backed so we don't lose that information

PR #12, opened 2016-09-04
Toggle to switch between asm2wasm instructions and unstable LLVM backend unstructions

s3ththompson:
I added a toggle to switch between the recommended asm2wasm instructions and the unstable LLVM backend instructions.

Live demo: https://rawgit.com/s3ththompson/webassembly.github.io/master/stepbystep/index.html

Does this look like an okay approach? @kripken @RebeccaRGBA

### DIFF
--- a/css/custom.css
+++ b/css/custom.css
@@ -10,14 +10,20 @@ ul {
   margin-left: 1.5em;
 }
 
-pre {
-  font-size: .9em;
-  padding: 1em;
-}
 code, kbd, pre, samp {
   font-family: Menlo,Monaco,Consolas,"Liberation Mono","Courier New",monospace;
   background-color: #f7f7f9;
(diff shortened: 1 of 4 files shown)


kripken:
Can you please elaborate on the motivation here? I'm not sure what this is aiming to do.

And more specifically, do you want people to know the difference between the two backends, and if so why? My feeling is that this adds unnecessary complexity, but maybe I'm missing an upside you see.

s3ththompson:
Good point on whether people should know the difference or not (ideally not). I think the motivation for this PR was basically just to provide more information for people interested in trying out the wasm backend. 

I'm happy with not merging this until either the backed is more stable or until it's ready to replace asm2wasm (in which case we wouldn't need a toggle at all).

PR #13, opened 2016-09-07
Add a version of Angrybots with most of the 0xc changes.

sunfishcode:
This adds 0xc changes, except block signatures.

Changes:

 - converted loops to not use their exit label
 - updated indirect_call operand order to put the callee last
 - inserted a gratuitous dependence on stack-machine rules (see the very
   first set_local)
 - inserted a gratuitous use of branch-to-function scope
 - optimized away unnecessary `blocks`
 - new "flat" text format


s3ththompson:
Do you want me to update the instantiation code to check for 0xC support and use this? Or should we wait to merge this until block signatures are added?

sunfishcode:
I think we should wait until block signatures are in. I posted this now because people expressed interested in it, but I hope to find time to add block signatures before long.

sunfishcode:
This PR is now subsumed by https://github.com/WebAssembly/webassembly.github.io/pull/14 which includes block signatures and also section opcodes, and also a candidate binary image.

PR #14, opened 2016-09-09
Candidate for 0xc AngryBots

sunfishcode:
Update to the remaining anticipated 0xc features.
    
Additional Changes:
 - Block signatures.
 - Section opcodes.
 - Minimally update Release/AngryBots.js to work with the new JS API.
 - Include a binary-encoded wasm file in 0xc format.
    
I made up a syntax for block signatures in the flat text format, taking
care to allow for adding arguments and multiple results in the future.
For example:
    
``` 
      block $label28 () : (i32)
``` 
    
Is a `block` no arguments and one return type which is i32.
    
However, I'm not attached to this particular syntax, and it can easily
be converted to something else if desired.

### DIFF
--- a/demo/AngryBots/Release/AngryBots.js
+++ b/demo/AngryBots/Release/AngryBots.js
@@ -97,7 +97,7 @@ function integrateWasmJS(Module) {
   };
   info["global.Math"] = global.Math;
   info["env"] = env;
-  if (Wasm.experimentalVersion < 0xc) {
+  if (Wasm.experimentalVersion < 0xc || typeof WebAssembly == "undefined") {
     var instance;
     instance = Wasm.instantiateModule(binary, info).exports;
     mergeMemory(instance.memory);
--- a/demo/AngryBots/Release/AngryBots.js
+++ b/demo/AngryBots/Release/AngryBots.js
@@ -98,14 +98,12 @@ function integrateWasmJS(Module) {
   info["global.Math"] = global.Math;
   info["env"] = env;
   if (Wasm.experimentalVersion < 0xc || typeof WebAssembly == "undefined") {
-    var instance;
-    instance = Wasm.instantiateModule(binary, info).exports;
-    mergeMemory(instance.memory);
+    var exports = Wasm.instantiateModule(binary, info).exports;
+    mergeMemory(exports.memory);
     applyMappedGlobals();
-    return instance;
+    return exports;
   }
-  var exports;
-  exports = new WebAssembly.Instance(new WebAssembly.Module(binary), info).exports;
+  var exports = new WebAssembly.Instance(new WebAssembly.Module(binary), info).exports;
   mergeMemory(exports.memory.buffer);
   applyMappedGlobals();
   return exports;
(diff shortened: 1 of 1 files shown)


sunfishcode:
The .wast file is now updated to use the block signature syntax of https://github.com/WebAssembly/spec/issues/336.

s3ththompson:
This won't run until we add polyfill support, right?

sunfishcode:
The index.html file uses `Wasm.experimentalVersion` to determine the filename to load. I've now renamed the wasm file to match the name it expects, so this should run in browsers that support the format.

s3ththompson:
Makes sense. Let me know when you're finished updating the binary and I'll lgtm

sunfishcode:
I've finished all the updates I'm aware of at this time.

The binary runs in my development tree of Firefox, which has all the anticipated 0xc binary format changes I'm currently aware of implemented.

s3ththompson:
Got it. Weren't we going to switch over to the `WebAssembly` object, though?

sunfishcode:
I've now updated the Angrybots.js script to switch to the new JS API, keeping the old Wasm.instantiateModule path in place when Wasm.experimentalVersion is less than 0xc.

demo/AngryBots/Release/AngryBots.js
```diff
@@ -97,11 +97,18 @@ function integrateWasmJS(Module) {
   };
   info["global.Math"] = global.Math;
   info["env"] = env;
-  var instance;
-  instance = Wasm.instantiateModule(binary, info).exports;
-  mergeMemory(instance.memory.buffer);
+  if (Wasm.experimentalVersion < 0xc) {
```
s3ththompson: perhaps `if (Wasm.experimentalVersion < 0xc || typeof WebAssembly == "undefined")` would be more robust (in case some engines haven't yet switched over)?

s3ththompson:
LGTM modulo minor nit

sunfishcode:
Nit fixed.

titzer:
I think we might have jumped the gun on landing this, as we haven't yet tried out these bits in V8.

s3ththompson:
@titzer Sorry, I didn't check yet because we haven't revved `Wasm.experimentalVersion` in Canary and I figured we can always update this binary if we find a problem.

lukewagner:
Well, almost: we were about to land our 0xc patch which, after landing, would make updating this binary more of a hassle.

@titzer Any ETA on V8 confirmation of the binary?

jfbastien:
Right, I think this was merged too quickly: we should be more disciplined about releasing binaries which we may just break in the future. No big deal, but let's not to that again.

titzer:
We should probably discuss how long we want browsers to support 0xC (and thus only demo bits) until the toolchain and everything else is also generating 0xC. I think we need a day or two to verify this binary and the other spec tests.

s3ththompson:
Sorry folks. Thought this was less of an issue given that the codepath won't be hit until we all rev `Wasm.experimentalVersion`, but I'll revert now for further discussion. Lesson learned.

s3ththompson:
Looks like I need an LGTM on the new revert PR

lukewagner:
@titzer That's a good point.  Having the binary verified will be a good first step, though, in demonstrating we're all on the same page.

s3ththompson:
Sorry for the churn. @sunfishcode would you mind rebasing your changes against the rebase and re-opening?

sunfishcode:
Rebased and reopened in #16.

PR #16, opened 2016-09-14
Candidate for 0xc AngryBots

sunfishcode:
This is the same change as https://github.com/WebAssembly/webassembly.github.io/pull/14.

Note that it includes the changes described in https://github.com/WebAssembly/design/pull/740
however it doesn't have any user-defined or names sections, so it's
unaffected by the details of how those are handled.


titzer:
lgtm (v8 0xC patch can grok these bits now)

lukewagner:
let's do this thing

PR #28, opened 2016-10-05
Update AngryBots-0xc.wast to be spec-compliant

binji:
Imports and exports syntax has changed:

    (import "module" "field" (func $name ...))
    (export "name" (func $name))
    (export "name" (memory 0))

Also the table syntax:

    (table anyfunc (elem ...))


bnjbvr:
LGTM. I've got a few patches to introduce the new syntax changes to Spidermonkey which make this file parse under our VM, and that would be great to wait for these to land in our CI before merging this (which should hopefully happen at the start of next week; I'll reping here once they have). Maybe other implementers are interested in getting some time too?

titzer:
V8 doesn't parse the WAST, on the binary. Has anyone tried this against the ml-proto's WAST parser?

bnjbvr:
I just tried using the ml-proto wast parser on the 0xc branch, and it does parse, since we're getting to the point where we have a link failure (`379.3-379.76: link failure: unknown import "global.Math.pow"`).

binji:
@bnjbvr status on the syntax changes in SM?

lukewagner:
It might be past his timezone so I'll confirm we're good here.

Issue #27, opened 2016-10-05
Update feature-detection code on demo


s3ththompson:
Proposal:

```js
if (typeof Wasm !== "undefined" && Wasm.experimentalVersion < 0xc) {
    var exports = Wasm.instantiateModule(binary, info).exports;
    mergeMemory(exports.memory);
    applyMappedGlobals();
    return exports;
} else if (typeof WebAssembly !== "undefined") {
  var exports = new WebAssembly.Instance(new WebAssembly.Module(binary), info).exports;
  mergeMemory(exports.memory.buffer);
  applyMappedGlobals();
  return exports;
} else {
  // asm.js fallback
}
```

lukewagner:
lgtm

titzer:
Won't we also have to do version probing to distinguish 0xC, 0xD, and 0x1?

lukewagner:
I was assuming this would be added (to the part where we compute the URL to fetch) with 0xD.

titzer:
SGTM.

s3ththompson:
I'll land this after #35 is finished.

lukewagner:
Included in #37.

PR #37, opened 2016-10-24
Update demo to 0xD and remove use of Wasm object

lukewagner:
This PR updates the demo to match SM's impl of the 0xD changes on `binary_0xd` branch in the design repo.  It also removes dependence on the `Wasm` object, using `WebAssembly.validate` to detect 0xc vs. 0xd, as discussed in #27.

Feedback welcome of whether this binary can be successfully decoded by other 0xD impls.

### DIFF
--- a/demo/AngryBots/index.html
+++ b/demo/AngryBots/index.html
@@ -95,15 +95,21 @@ window.addEventListener('resize', softFullscreenResizeWebGLRenderTarget);
     codeUrl: "Release/AngryBots.js",
     memUrl: "Release/AngryBots.mem",
     compatibilitycheck: function() {
-      if (typeof Wasm !== 'object') (alert("You need a browser which supports WebAssembly to run this content. See the WebAssembly demo page for instructions."), window.history.back());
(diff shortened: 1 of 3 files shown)


demo/AngryBots/index.html
```diff
@@ -95,15 +95,21 @@
     codeUrl: "Release/AngryBots.js",
     memUrl: "Release/AngryBots.mem",
     compatibilitycheck: function() {
-      if (typeof Wasm !== 'object') (alert("You need a browser which supports WebAssembly to run this content. See the WebAssembly demo page for instructions."), window.history.back());
+      if (typeof Wasm !== 'object' && typeof WebAssembly !== 'object') (alert("You need a browser which supports WebAssembly to run this content. See the WebAssembly demo page for instructions."), window.history.back());
```
jfbastien: Would it be better to redirect to webassembly.org instead of history back?

lukewagner: I suppose we could; I'm mostly just focusing on 0xD here, though.

titzer:
LGTM. Confirmed this works in Chrome with 0xD patch applied.

PR #38, opened 2016-10-26
use WebAssembly.compile in demo, if present

lukewagner:
This PR feature-tests and uses `WebAssembly.compile`, if present, so that the demo can take advantage of parallel compilation.  It also adds some comments for anyone reading the code to understand what bits are temporary hacks.

### DIFF
--- a/demo/AngryBots/Release/UnityLoader.js
+++ b/demo/AngryBots/Release/UnityLoader.js
@@ -82,6 +82,12 @@ function LoadCompressedJS(e, t) {
     })
 }
 
+function LoadUncompressedJS(e, t) {
+    var script = document.createElement('script');
+    script.src = e;
+    document.body.append(script);
+}
+
 function fetchRemotePackageWrapper(e, t, n, o) {
     LoadCompressedFile(e, n, function(n) {
         var o = e,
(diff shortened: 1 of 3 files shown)


demo/AngryBots/Release/UnityLoader.js
```diff
@@ -158,25 +164,49 @@ var CompressionState = {
...
+
+// For simplicity, .wasm files are not currently compressed but in a production
+// build, they would be (and use LoadCompressedFile).
+(function() {
+    var xhr = new XMLHttpRequest();
+    xhr.open('GET', Module.wasmUrl, true);
+    xhr.responseType = 'arraybuffer';
+    xhr.onload = function() {
+        if (typeof Wasm == 'undefined' && WebAssembly.compile) {
+            WebAssembly.compile(xhr.response).then(function(module) {
```
jfbastien: @MikeHolman will that work on IE? `.compile` could be present without `.then` or is that not the case anymore?

lukewagner:
@jfbastien Well aware that those are different concepts; I do mean parallelism.  And extended sentence could have ended "... when present in the browser's implementation of `compile`".  Asynchrony without the parallelism (so the compilation happens on the main thread at some unspecified future turn of the event loop) isn't much of a win.

demo/AngryBots/Release/UnityLoader.js
lukewagner: `WebAssembly.compile` was defined to return a `Promise` from day 1; are you confusing this with `Wasm.instantiateModule`?

jfbastien:
I mean the other way around: `WebAssembly.Module` can totally be parallel but not be synchronous.

demo/AngryBots/Release/UnityLoader.js
jfbastien: I was rather asking if `.then` works, but I see that I read IE as not supporting it whereas Edge does.

s3ththompson:
ping on this, @lukewagner is this ready to merge? @jfbastien want to lgtm?

jfbastien:
Sure, lgtm.

PR #47, opened 2016-11-02
Update demo instructions: Chrome now supported

mheiber:
Chrome canary is no longer needed, the
`chrome://flags/#enable-webassembly`
flag works in Chrome.


jiyinyiyong:
I'm using Canary but turned out I still need to turn it on by myself in "Version 56.0.2906.0 canary (64-bit)".

jfbastien:
I think only Canary support version `0xD` though, right @titzer / @s3ththompson ?
The demo will work because it tests for various versions, but it would be misleading to have people try to build their own thing and see it fail because they're not using canary. For that reason, I think it's better to not edit the instructions.

s3ththompson:
+1, agreed on (for now) having people continue to use Canary.

jfbastien:
OK, closing for now.

PR #56, opened 2016-11-23
Change quites in the code examples

yurydelendik:
That helps to avoid confusion and compile errors when the sample code is copied "as is" from the page.


lukewagner:
Hah, good one.

PR #57, opened 2016-11-28
Update site with small fixes & recent design repo commits

s3ththompson:
- typo on JS API page
- new V8 bug link
- update to HEAD of design repo


flagxor:
looks like this pulls in fresher versions of the other repos too.
lgtm

PR #58, opened 2016-12-12
Update docs

franckverrot:
Here are two mini updates to make sure the example compiles properly.

Thanks for all the amazing work!

### DIFF
--- a/docs/getting-started/developers-guide/index.html
+++ b/docs/getting-started/developers-guide/index.html
@@ -79,7 +79,7 @@ $ mkdir hello
 $ cd hello
 $ echo '#include &lt;stdio.h&gt;' &gt; hello.c
 $ echo 'int main(int argc, char ** argv) {' &gt;&gt; hello.c
-$ echo 'printf("Hello, world!\n");' &gt;&gt; hello.c
+$ echo 'printf("Hello, world!\\n");' &gt;&gt; hello.c
 $ echo '}' &gt;&gt; hello.c
 $ <b>emcc hello.c -s WASM=1 -o hello.html</b>
 </pre>
--- a/getting-started/developers-guide.md
+++ b/getting-started/developers-guide.md
@@ -37,7 +37,7 @@ $ mkdir hello
 $ cd hello
 $ echo '#include &lt;stdio.h&gt;' &gt; hello.c
 $ echo 'int main(int argc, char ** argv) {' &gt;&gt; hello.c
-$ echo 'printf("Hello, world!\n");' &gt;&gt; hello.c
+$ echo 'printf("Hello, world!\\n");' &gt;&gt; hello.c
 $ echo '}' &gt;&gt; hello.c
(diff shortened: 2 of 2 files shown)


franckverrot:
Hi @kripken,

Yes, I get this:

```
λ emcc hello.c -s WASM=1  -o hello.html
build executables
save the before js
process input
cannot use WASM=1 when full asm.js validation was disabled (make sure to run in at least -O1, and look for warnings about other options that might force asm.js off)
```

kripken:
That sounds familiar, I think we had it a while back. What emscripten version are you on? (`emcc --version`)

franckverrot:
```
λ emcc --version
emcc (Emscripten gcc/clang-like replacement) 1.35.0 ()
Copyright (C) 2014 the Emscripten authors (see AUTHORS.txt)
This is free and open source software under the MIT license.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

kripken:
Looks like emscripten's master branch, which is is somewhat behind. The docs give instructions that should install incoming (1.36.14). How did you install emcc, maybe something went wrong or you have more than one version installed?

franckverrot:
My bad, I used `emsdk install latest` thinking this would be the latest but defaulted to 1.35.0.

It's working without now.

franckverrot:
Just pushed an update on this branch to get the original source file right.

kripken:
Hmm, for me on linux entering `echo 'printf("Hello, world!\n");'` works fine, the way the site is now. If I add another `\` as you suggest then I see 2 slashes in the source file, which isn't right. Do you see something different?

franckverrot:
I'm running zsh (on OSX); bash and ksh behave OK though (and I didn't check that before triggering the PR, apologies... ).  Closing this PR (as the only thing I could think of to improve the doc would be to mention some `cat >> hello.c << EOF...`), thanks for all the help! 😄

kripken:
No problem. Maybe we should mention the zsh escaping thing though? I don't know much about zsh (what's the default on OSX?) but if it's common then seems like that would be helpful to mention?

franckverrot:
It's not the default (bash is). 

I can either try replacing the whole block with something like:

```
mkdir hello
cd hello
cat >> hello.c << EOF
#include <stdio.h>
int main(int argc, char ** argv) {
  printf("Hello, world!\n")
}
EOF
emcc hello.c -s WASM=1 -o hello.html
```

or just add a warning like:
 __If you are running zsh, be sure to run `setopt BSD_ECHO` first as zsh will [escape sequences in echo-ed strings](http://zsh.sourceforge.net/Doc/Release/Options.html)__.

Do you have any preference? Still think it's worth mentioning?

kripken:
Hmm, if zsh isn't the default, then maybe not worth it (to avoid clutter, there are probably lots of other non-default options that require changes).

PR #59, opened 2016-12-19
Add Node talk to community events page

linclark:
I'm not sure whether the scope of this page is limited to talks given by people working on the standard. If it is not, this is a helpful talk that demonstrates using WASM with Node. The Node workflow is still clunky, but I think adventuresome developers might want to tinker with it, and this is the only demo of it I've seen.


jfbastien:
I'm watching it now, it's a pretty great talk! We haven't really thought about what to include as more and more talks happen, but I think high-quality talks should be here.

@lukewagner / @titzer thoughts?

lukewagner:
Yeah, definitely lgtm, thanks!

PR #64, opened 2017-01-05
Update getting started developer's guide.

juj:
Improve getting started instructions to refer to the appropriate steps for building latest Emscripten from source. Also use `emrun` instead of python web server, because `emrun` knows to provide proper HTTP response headers to the served files, which SimpleHTTPServer is not aware of. Add Windows instructions as well, and a section on prerequisites.

Once precompiled Emscripten packages are available which don't need building from source, the getting started docs can be rewritten to not require git, CMake or Visual Studio/gcc/Xcode.

### DIFF
--- a/getting-started/developers-guide.md
+++ b/getting-started/developers-guide.md
@@ -4,30 +4,39 @@ layout: getting-started
 
 # Developer's Guide
 
-This page aims to provide complete, working step-by-step instructions to compile a simple program directly to WebAssembly.
+This page provides step-by-step instructions to compile a simple program directly to WebAssembly.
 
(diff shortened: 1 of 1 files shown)


kripken:
lgtm

s3ththompson:
lgtm

jfbastien:
There should be a pre-commit hook for this 😛

juj:
Tried rebuilding the site, I see I've got Ruby 2.0, and then proceeded to the next step of "install gem dependencies", but I see I don't have the "bundle" executable available on step "bundle install", not sure what that is? I wonder if I'm missing some dependency?

s3ththompson:
Ah right, will document installing http://bundler.io/.

In the meantime, no worries, will build site while merging.

PR #66, opened 2017-01-09
fix demo file extension

chicoxyzzy:


sunfishcode:
In some circles, ".wast" indicates conformance test files, which have turned out to use a different syntax than ".was" which indicates wasm's [anticipated official text format](https://github.com/WebAssembly/design/blob/master/TextFormat.md#linear-bytecode). However, this distinction is not yet widely reflected in documentation or tools.

PR #71, opened 2017-02-06
Remove chrome canary instruction about flag

thejameskyle:
Latest Chrome Canary has wasm enabled by default


s3ththompson:
Thanks for the PR. This was removed in #78

PR #74, opened 2017-02-17
Adds warning re:browser preview to getting-started

AVGP:
The current `incoming` branch produces wasm binaries with version `0x1` instead of the expected `0xd`.
This can be resolved by using the `master` branch instead.


s3ththompson:
This should be resolved with the latest `incoming` changes. Thanks for opening this though.

PR #80, opened 2017-03-04
Changed to 'emsdk_env.bat' from 'emsdk_env'

achung89:
I believe the windows command to add emsdk to PATH is 'emsdk_env.bat' [link to section in documentation](https://github.com/juj/emsdk#installing-emsdk-directly-from-github)


kripken:
cc @juj

juj:
On Windows one doesn't need to suffix executable files with the extension, so it's alright to just call `emsdk_env` on Windows instead of `emsdk_env.bat`, though certainly no harm to merge this in if there's a preference toward having the extension visible in the docs?

PR #95, opened 2017-05-30
Add the logo to the website

gahaas:
This PR adds the WebAssembly logo from https://github.com/carlosbaraza/web-assembly-logo to the website. For this I added the logo to css/webassembly.svg and changed the files _includes/header.html and css/custom.css. All other files were generated automatically.


s3ththompson:
@gahaas thanks for the PR! I think it would be better to just use the WebAssembly icon, and keep the title as text, so that we can display a useful title on mobile.

WDYT of this? (we might need to make the buttons a little smaller)

screen shot 2017-05-30 at 1 50 56 pm
screen shot 2017-05-30 at 1 51 28 pm

gahaas:
I just added the logo with the text now also to mobile, which looks good in my opinion. I think it is good to use the logo with text, as the text looks nicer, and we do not have to play so many css tricks to keep stuff aligned. Also the buttons don't have to be smaller if we do not put the logo next to the text but on top of it.

s3ththompson:
makes sense. lgtm

PR #100, opened 2017-07-07
Add link to spec draft

rossberg-chromium:
...to make it more easily discoverable.


s3ththompson:
Can you fix CSS so that navigation bar doesn't overflow?

screen shot 2017-07-13 at 4 05 48 pm

cc @thomasnat1 will take over from here!

thomasnat1:
I've add the change you suggested to all the pages and changed the CSS slightly so they fit on one line.

PR #104, opened 2017-08-16
getting_started_with_precompiled_toolchain

juj:
We have had precompiled toolchains distributed for a while now, so update Getting Started documentation to include a section on obtaining a precompiled toolchain.

PR #107, opened 2017-09-27
Update and simplify getting-started docs

kripken:
In particular the emsdk does not build from source if it doesn't need to, any more.


kripken:
cc @juj

juj:
This is a duplicate of pending #104 it seems. Let me ping there again.

thomasnat1:
Closing since #104 was merged

kripken:
Perhaps we still want parts of this PR? Two things it adds over the current master are

 * Mention that you can download the emsdk, you don't need to git clone it (might be faster).
 * Links to the emsdk docs (which might be more up to date at times).

PR #109, opened 2017-12-22
Update Rationale.md

sherhut:
Remove superfluous occurrence of word 'model'.


jfbastien:
This site mirrors the sources from: https://github.com/WebAssembly/design/blob/master/Rationale.md

I've fixed the issue at the origin. Thanks!

PR #112, opened 2018-01-26
Simple update for "hello world" commands

rfdavid:
Just a more cleaner commands set for the "hello world" example.


muhammadmuzzammil1998:
I concur. 👍

PR #115, opened 2018-04-03
Update lock-file to match Gemfile

LinusU:
ce9efaa7a475d9b4d0228bcb553ee1abad10a487 updated the Gemfile but not the lock-file which caused installs to fail, easy fix 👌


LinusU:
Seems like this was fixed by #117

PR #116, opened 2018-04-03
Add higher resolution of Safari logo

LinusU:
The top one is after, the bottom one before (on my 2x retina screen)

screen shot 2018-04-03 at 17 24 33


LinusU:
Thanks!

PR #117, opened 2018-05-03
Change http:// to https:// in URLs that support HTTPS

macsual:
All the modified URLs were checked to see if they are available over HTTPS by opening a new tab in the browser, entering the URL in the address bar, changing `http://` to `https://` and then verifying if the resource displayed is the same over HTTP and HTTPS for each URL.

[asmjs.org](http://asmjs.org/) and [emscripten.org](http://emscripten.org/) are [custom domains](https://help.github.com/articles/using-a-custom-domain-with-github-pages/) for [GitHub Pages sites](https://pages.github.com/) which [can now become available over HTTPS](https://blog.github.com/2018-05-01-github-pages-custom-domains-https/) but action is required on each domain's owner's part to secure these custom domains before the URLs may be rewritten to change `http://` to `https://`.


jfbastien:
Travis seems sad :(

Issue #122, opened 2018-08-22
Incorrect link to docs when user is on developer guide page


mathiashoeld:
On https://webassembly.org/ after navigating to the developer's guide (https://webassembly.org/getting-started/developers-guide/) the link to "Docs" is incorrect. It points to the same page as the developer guide. 

Windows 10 using Chrome Version 68.0.3440.106

binji:
This seems to be fixed now, please reopen if not.

PR #123, opened 2018-09-11
Fix broken link

ollelauribostrom:
Associated Issue: #122 

Commit 5a00907 seems to have left the href in the docs link empty (on the developer's guide page). This PR adds it back again.


binji:
Thanks! Sorry for the delay. I just went through and re-generated the docs which seems to have fixed this.

PR #129, opened 2019-01-21
Update links to emsdk repo, which has been transfered to emscripten-core/

kripken:
(Things do work in the meantime since github magically redirects everything.)


binji:
I rebased this patch, seems good to me. Travis is failing here because it was testing w/ newer bundles on ruby 2.2, but I updated travis to ruby 2.3 so this should be fine now.

PR #130, opened 2019-01-29
Add Go language to the index page.

gios:
The [**Go team**](https://github.com/golang/go/wiki/WebAssembly) has made additional changes to the compiler for compiling Go programs to wasm. We should add this language to the main page.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -19,8 +19,8 @@
       <a class="site-nav-item btn" href="/">Overview</a>
       <a class="site-nav-item btn" href="/demo/">Demo</a>
       <a class="site-nav-item btn" href="/getting-started/developers-guide/">Getting Started</a>
-      
-      
+
+
       <a class="site-nav-item btn" href="/docs/high-level-goals/">Docs</a>
       <a class="site-nav-item btn" href="https://webassembly.github.io/spec/">Spec</a>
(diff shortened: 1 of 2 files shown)


rossberg:
To be honest, I'd rather suggest removing any mention of specific languages, since we're already beyond just a few, and it would be silly to enumerate them all.

titzer:
I'd suggest we link to a separate page which lists languages and their implementation/support status. Such a page is pretty useful and would be good advertising for Wasm.

rossberg:
Do you mean a page hosted by us? There would be some things to consider:

* There are external sites hosting such lists already, e.g. [here](https://github.com/appcypher/awesome-wasm-langs). Should we duplicate these efforts?

* Do we want to get into the politics of deciding what's a proper language implementation and what just somebodies toy compiler?

* Such a list might cause more harm than good if we don't maintain it continuously. Somebody would need to commit to owning that.

RReverser:
> I'd suggest we link to a separate page which lists languages and their implementation/support status. Such a page is pretty useful and would be good advertising for Wasm.

This is now basically done as part of #161: https://webassembly.org/getting-started/developers-guide/

If we want to do more, I think it would be better to start a separate discussion.

RReverser:
Closing in favour of #165.

PR #136, opened 2019-03-17
Link to official LLVM repo instead

dduan:
LLVM has an official repository on Github now. Link to it instead of the mirror
repo.


binji:
Thanks!

Issue #143, opened 2019-07-09
Nav bar SVG does not link


shinmem58:
The SVG icon in the nav bar does not link to the main page as it usually is in websites.
Is it on purpose?

binji:
No, but that seems like a good fix. We're planning on doing a webassembly.org redesign soon, but if you'd like to submit a fix for this, we can land it in the meantime.

shinmem58:
Thank you

PR #144, opened 2019-07-10
Aded SVG link to home page

shinmem58:
As discussed in issue #143  I added to the SVG a link to the home page.
I added href and xlink:href for backwards compatibility and for safari.

For some reason when I compare the code it shows that I rewrote the SVG when a really added lines 14, 36.

### DIFF
--- a/css/webassembly.svg
+++ b/css/webassembly.svg
@@ -11,6 +11,7 @@
     </style>
   </defs>
   <title>web-assembly-logo</title>
+  <a href="/" xlink:href="/">
   <g id="Layer_2" data-name="Layer 2">
     <g id="Notch_-_Purple" data-name="Notch - Purple">
       <g>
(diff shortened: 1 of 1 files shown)


binji:
I tried loading this SVG in Chrome and it gives the following error:

```
This page contains the following errors:
error on line 14 at column 29: Namespace prefix xlink for href on a is not defined
Below is a rendering of the page up to the first error.
```

shinmem58:
I fixed the problem.
Should I make a new pull rquest ?

RReverser:
What was this change about? It's hard to tell with all the indentation changes.

RReverser:
Ah I see now on the original issue. However, I'd prefer to add a link in HTML instead - it should be as easy as changing `div` to `a`.

RReverser:
I've tried to use this PR, but it doesn't work because SVG is used in `background-image` context where it can't have links. I'll switch `div` -> `a` as described above instead.

PR #150, opened 2019-09-24
Attempt at fixing CI

pygy:
This is a blind try at fixing Travis, based on their [docs](https://docs.travis-ci.com/user/languages/ruby/#bundler-20). Let's see if CI passes with this change...


kripken:
Not sure who should review this, but it does fix CI, so that sounds great, thanks @pygy!

PR #158, opened 2020-03-31
Minimized SVG files

FlipFloop:


RReverser:
These are the source files; we automatically minify them for production as part of the build step.

RReverser:
I was wrong. We should, but looks like we don't :(

tomayac:
Closing in favor of #416.

PR #161, opened 2020-05-27
Update Getting Started page

RReverser:
*👋🏻 As some of you know, I've recently taken over work of updating the website to include more relevant content for new developers in this new age. This is the first in a series of small PRs towards that goal.*

Instead of duplicating an Emscripten tutorial, provide a curated list of links to tutorials or documentation that helps people get started with WebAssembly from various languages and environments.

This is an initial list and can and will be expanded over time.

### DIFF
--- a/getting-started/developers-guide.md
+++ b/getting-started/developers-guide.md
@@ -2,65 +2,31 @@
 layout: getting-started
 ---
 
-# Developer's Guide
+# I want to...
 
-This page provides step-by-step instructions to compile a simple program directly to WebAssembly.
+- Compile a WebAssembly module from...
+  - C/C++
+    - [starting from scratch](https://developer.mozilla.org/en-US/docs/WebAssembly/C_to_wasm)
+    - [library that I want to port to the Web](https://developer.mozilla.org/en-US/docs/WebAssembly/existing_C_to_wasm)
+  - [Rust](https://developer.mozilla.org/en-US/docs/WebAssembly/Rust_to_wasm)
+  - [AssemblyScript (a TypeScript-like syntax)](https://docs.assemblyscript.org/)
+  - [C#](https://docs.microsoft.com/en-us/aspnet/core/blazor/get-started?view=aspnetcore-3.1&tabs=visual-studio)
+  - [F#](https://fsbolero.io/docs/)
+  - Go
+    - [with full language support](https://github.com/golang/go/wiki/WebAssembly#getting-started)
+    - [targeting minimal size](https://tinygo.org/webassembly/webassembly/)
+  - [Kotlin](https://kotlinlang.org/docs/reference/native-overview.html)
+  - [Swift](https://swiftwasm.org/)
+  - [D](https://wiki.dlang.org/Generating_WebAssembly_with_LDC)
+  - [Pascal](https://wiki.freepascal.org/WebAssembly/Compiler)
+- Use the compiled WebAssembly...
+  - [from JavaScript code](https://developer.mozilla.org/en-US/docs/WebAssembly/Loading_and_running)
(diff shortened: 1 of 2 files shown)


getting-started/developers-guide.md
```diff
@@ -2,65 +2,31 @@
...
-
-    $ emrun --no_browser --port 8080 .
-
-Once the HTTP server is running, you can <a href="http://localhost:8080/hello.html" target="_blank">open it in your browser</a>. If you see "Hello, world!" printed to the Emscripten console, then congratulations! You've successfully compiled to WebAssembly!
+# I want to...
+
+- Compile a WebAssembly module from...
+  - C/C++
+    - [starting from scratch](https://developer.mozilla.org/en-US/docs/WebAssembly/C_to_wasm)
+    - [library that I want to port to the Web](https://developer.mozilla.org/en-US/docs/WebAssembly/existing_C_to_wasm)
```
kripken, line 10: These 2 links are very good but I'd recommend also having the Emscripten tutorial, https://emscripten.org/docs/getting_started/Tutorial.html

RReverser: That's something I wanted to discuss separately (or I think mentioned already before?) - I believe that emscripten.org, especially the tutorial, should be updated before we can link out to it from this page.

It's what I wanted to link to initially, but, unfortunately, I think that tutorial is rather advanced and might be overwhelming for beginners in its current state with the jumps right away from "hello world" to SDL usage and more.

In this spirit, MDN links seem more tailored for a gettin started guide, as it focuses on most basic concepts (compiling, running and exposing a function) without going into advanced stuff.

kripken: Oh, I see. Happy to discuss that separately if you don't want the noise in this PR. Also I'd be very happy to get suggestions and help for improving that page!

One reason I think an `emscripten.org` link is a good idea is that we update it together with the rest of our docs with all of emscripten. Do we know who updates those MDN links?

RReverser: I think they're maintained by community in general - since MDN is a wiki, it's easy for anyone to contribute.

This can be seen as both pros and cons at the same time, but we want to move webassembly.org to be a more community-oriented website, so maybe it's a good thing?

Anyway, happy to discuss separately as well, and, as mentioned in PR description, we can always extend / update the list, for now I just want to have a best-effort attempt at a list of getting started guides that beginners can jump into. When emscripten.org updates, it should be easy to switch over.

getting-started/developers-guide.md
```diff
@@ -2,65 +2,31 @@
...
+  - [Rust](https://developer.mozilla.org/en-US/docs/WebAssembly/Rust_to_wasm)
+  - [AssemblyScript (a TypeScript-like syntax)](https://docs.assemblyscript.org/)
+  - [C#](https://docs.microsoft.com/en-us/aspnet/core/blazor/get-started?view=aspnetcore-3.1&tabs=visual-studio)
+  - [F#](https://fsbolero.io/docs/)
+  - Go
+    - [with full language support](https://github.com/golang/go/wiki/WebAssembly#getting-started)
+    - [targeting minimal size](https://tinygo.org/webassembly/webassembly/)
+  - [Kotlin](https://kotlinlang.org/docs/reference/native-overview.html)
```
binji, line 18: Is there a more wasm-specific link we can point to here?

RReverser: I used the best I could find (it seems Kotlin/Native is sharing same official docs across supported targets), but if there is, I'm happy to change.

RReverser: (I found some 3rd-party blog posts but not feeling very comfortable linking out to them)

binji: OK, the link you have is pretty good for now. We can always update later if we find a better one.

PR #163, opened 2020-05-27
Add Specifications page

RReverser:
*👋🏻 As some of you know, I've recently taken over work of updating the website to include more relevant content for new developers in this new age. This is one of the first in a series of small PRs towards that goal.*

Instead of linking out to an unstyled https://webassembly.github.io/spec/ page only, provide a list of links to various specs related to WebAssembly in our own UI.

### DIFF
--- a/specs.md
+++ b/specs.md
@@ -3,7 +3,7 @@ layout: default
 ---
 # Specifications
 
-- [Design documents](https://github.com/WebAssembly/design): documents describing the design, goals and high-level overview of WebAssembly.
+- [Original design documents](https://github.com/WebAssembly/design): documents describing the design, goals and high-level overview of WebAssembly. Some of these documents are outdated by now.
 - [Core specification](https://webassembly.github.io/spec/core/): defines the semantics of WebAssembly modules independent from a concrete embedding. The WebAssembly core is specified in a single document.
 - Embedding interfaces:
   - [JavaScript API](https://webassembly.github.io/spec/js-api/index.html): defines JavaScript classes and objects for accessing WebAssembly from within JavaScript, including methods for validation, compilation, instantiation, and classes for representing and manipulating imports and exports as JavaScript objects.
--- a/specs.md
+++ b/specs.md
@@ -3,10 +3,10 @@ layout: default
 ---
 # Specifications
 
-- [Original design documents](https://github.com/WebAssembly/design): documents describing the design, goals and high-level overview of WebAssembly. Some of these documents are outdated by now.
 - [Core specification](https://webassembly.github.io/spec/core/): defines the semantics of WebAssembly modules independent from a concrete embedding. The WebAssembly core is specified in a single document.
 - Embedding interfaces:
   - [JavaScript API](https://webassembly.github.io/spec/js-api/index.html): defines JavaScript classes and objects for accessing WebAssembly from within JavaScript, including methods for validation, compilation, instantiation, and classes for representing and manipulating imports and exports as JavaScript objects.
   - [Web API](https://webassembly.github.io/spec/web-api/index.html): defines extensions to the JavaScript API made available specifically in web browsers, in particular, an interface for streaming compilation and instantiation from origin-bound `Response` types.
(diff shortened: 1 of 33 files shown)


specs.md
```diff
@@ -0,0 +1,12 @@
+---
+layout: default
+---
+# Specifications
+
+- [Design documents](https://github.com/WebAssembly/design): documents describing the design, goals and high-level overview of WebAssembly.
+- [Core specification](https://webassembly.github.io/spec/core/): defines the semantics of WebAssembly modules independent from a concrete embedding. The WebAssembly core is specified in a single document.
+- Embedding interfaces:
+  - [JavaScript API](https://webassembly.github.io/spec/js-api/index.html): defines JavaScript classes and objects for accessing WebAssembly from within JavaScript, including methods for validation, compilation, instantiation, and classes for representing and manipulating imports and exports as JavaScript objects.
```
binji, line 9: Some of the descriptions here have contents that could become out-of-date; I guess that's OK since we can always update them, but I wonder if it makes more sense to use a description here that doesn't really need to change.

RReverser: FWIW these descriptions are pretty much copy-pasted from https://webassembly.github.io/spec/ which we were pointing to before.

If they're out of date, it's probably worth updating them at the source as well, but I'm happy to accept any fixes here :)

binji: Ah, fair enough :) They're accurate enough for now, lets keep them as is.

specs.md
```diff
@@ -0,0 +1,12 @@
+---
+layout: default
+---
+# Specifications
```
rossberg, line 4: Strictly speaking, not all docs in this list are specifications. How about "Specifications and Design Documents"?

specs.md
```diff
@@ -0,0 +1,12 @@
...
+
+- [Design documents](https://github.com/WebAssembly/design): documents describing the design, goals and high-level overview of WebAssembly.
```
rossberg, line 6:
```suggestion
- [Original design documents](https://github.com/WebAssembly/design): documents describing the design, goals and high-level overview of WebAssembly. Same of these documents are outdated by now.
```

RReverser: Interesting; are they not relevant anymore for new additions?

specs.md
RReverser: Maybe, although I'm not sure we care about strict / detailed definition in the title.

Otherwise we'll need to update the top button which currently just says "Specs" too, and we don't have too much space there.

RReverser: I think for most users "Specifications" is a generic enough cover word to include design documents too; beyond that, it seems to be a nuanced detail which is already expressed in a separate list item.

specs.md
RReverser: Note that I've already updated things like "Binary encoding", "Text format" and "Semantics" to redirect to the up-to-date spec and was thinking about removing them from the upstream `design` repo to avoid confusion if possible.

However, I thought that most other docs still apply to WebAssembly today - if not, maybe we should work out which ones should be split out in their own folder or repo as "legacy" / "original".

rossberg: Most of the design docs haven't been updated in a long time. I can't tell off-hand what's still in sync, but expect there to be small deviations throughout. More importantly, these docs do not cover any of the new features, whose design docs are in the respective proposal repos.

FWIW, I don't think we should delete any of the original design docs. They are valuable as a documentation for the rationale behind the original MVP design. But we need to make clear that they are no longer an up-to-date reflection of the status quo.

RReverser: > FWIW, I don't think we should delete any of the original design docs. They are valuable as a documentation for the rationale behind the original MVP design. But we need to make clear that they are no longer an up-to-date reflection of the status quo.

Agreed, which is why I think that moving them into own folder (or a repo) might be the best of both worlds.

RReverser: I've raised another issue on the upstream repo for that. I think this change is fine, but when we separate up-to-date documents from historical ones, we should revise this item again.

RReverser: Merged & moved them to the end of the list, since they sound even more niche for most users than initially thought.

specs.md
RReverser: As mentioned in another thread, I moved design docs down in the list, so I think mentioning them in the title might be an unnecessary distraction.

PR #165, opened 2020-06-11
Remove specific langs from lead

RReverser:
We'll probably want to redesign / reword this page further, but for now this should suffice as a more generic alternative to #130.

### DIFF
--- a/index.md
+++ b/index.md
@@ -1,6 +1,6 @@
 ---
 layout: default
-lead: WebAssembly (abbreviated <i>Wasm</i>) is a binary instruction format for a stack-based virtual machine. Wasm is designed as a portable target for compilation of native languages, enabling deployment on the web for client and server applications.
+lead: WebAssembly (abbreviated <i>Wasm</i>) is a binary instruction format for a stack-based virtual machine. Wasm is designed as a portable compilation target for programming languages, enabling deployment on the web for client and server applications.
 ---
 <div class="flash flash-warn">
   Developer reference documentation for Wasm can be found on <a href="https://developer.mozilla.org/en-US/docs/WebAssembly">MDN's WebAssembly pages</a>.
--- a/index.md
+++ b/index.md
@@ -1,6 +1,6 @@
 ---
 layout: default
-lead: WebAssembly (abbreviated <i>Wasm</i>) is a binary instruction format for a stack-based virtual machine. Wasm is designed as a portable target for compilation of high-level languages like C/C++/Rust, enabling deployment on the web for client and server applications.
+lead: WebAssembly (abbreviated <i>Wasm</i>) is a binary instruction format for a stack-based virtual machine. Wasm is designed as a portable target for compilation of native languages, enabling deployment on the web for client and server applications.
 ---
 <div class="flash flash-warn">
   Developer reference documentation for Wasm can be found on <a href="https://developer.mozilla.org/en-US/docs/WebAssembly">MDN's WebAssembly pages</a>.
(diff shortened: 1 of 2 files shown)


index.md
```diff
@@ -1,6 +1,6 @@
 ---
 layout: default
-lead: WebAssembly (abbreviated <i>Wasm</i>) is a binary instruction format for a stack-based virtual machine. Wasm is designed as a portable target for compilation of high-level languages like C/C++/Rust, enabling deployment on the web for client and server applications.
+lead: WebAssembly (abbreviated <i>Wasm</i>) is a binary instruction format for a stack-based virtual machine. Wasm is designed as a portable target for compilation of native languages, enabling deployment on the web for client and server applications.
```
sbc100, line 3: Is the concept of a "native language" well understood?   I'm not sure I could define it.  Perhaps "compiled languages"?   Although I feel like the existing wording is still better.  It does say "like" so it clearly not exhaustive.   What are your specific concerns about the existing wording?

RReverser: See #130; basically, it makes some languages look more preferable than others, and then naturally folks from other languages want to see those listed as well.

Instead, I'd prefer for us to have a neutral description here, and, while "native" can be still ambiguous as you noted, it's generally well-understood in the context of "web" (which this sentence mentions further).

binji: I agree that "native" sounds odd. I'd be ok with leaving it as "high-level languages" or even "various high-level languages"

sbc100: How about ... `compilation of high-level languages (for example C, C++ and Rust)`.  Maybe using "for example" would sounds less biased?

Personally, I think that "high-level languages" is clearer than "native languages" .   Native languages makes me think of my own native language which is English.  But maybe thats just me? 

Out of interest, what is your understanding of the term native languages?    Does it include interpreted languages or only compiled ones?   It feels to me like it only includes compiled languages which maybe is fine?

Sorry to nit pick here..

RReverser: > I'd be ok with leaving it as "high-level languages"

Yeah but arguably that's an equally ambiguous term - few people consider C a "high-level language" for example, yet it was listed there. This way "native languages" seems broader in that it includes low-level native languages as well.

> Does it include interpreted languages or only compiled ones?

Yeah, just compiled ones. Maybe I'm the odd one, but I've seen that term used interchangeable with "compiled languages", and in this case "compilation of compiled languages" would sound very odd, so I went for a close replacement.

kripken: +1 for removing the specific list of languages, as there is no need to favor some over others.

+1 for avoiding "native" - I agree it's not a clear enough term, but also, just like a specific list it seems to favor some languages more than others, either "compile-to-native-binary ones" (which might seem to leave out, say, C#) or "non-web" (which might seem to leave out, say, AssemblyScript) or "non-GC" (but we intend to have wasm GC eventually), etc. Compile all the things to wasm!

How about just `Wasm is designed as a portable target for languages to compile to`?

RReverser: > which might seem to leave out, say, C#

I'd argue the statically compiled subset of C# is a native language, but I can see where this is going :)

> Wasm is designed as a portable target for languages to compile to

It sounds a bit odd to my non-native ear (well, not ear, but you know what I mean) when there is no adjective before "languages"...

Maybe something like "Wasm is designed as a portable compilation target for programming languages" would be sufficiently precise yet generic at the same time?

kripken: Not a native speaker either, but that sounds good to me.

RReverser: Updated.

binji: Works for me too, thanks!

PR #120, opened 2018-07-04
Update favicon with 2x resolution.

lgarron:
This updates the favicon to include a 2x resolution version for Retina screens.

I've also updated the 16px version to have a transparent puzzle piece indentation at the top (instead of white).

screen shot 2018-07-04 at 12 41 54

PR #166, opened 2020-06-15
Add zig to getting started

kubo39:


RReverser:
Thanks for your contribution! As you can see, the Travis has failed due to HTML mismatch - unfortunately, we currently commit all build artifacts, which means that you need to build website locally and update the PR. Could you please do that?

I think you also need to join WebAssembly CG, although I'm not sure how strict that requirement is for contributions. @binji ?

RReverser:
Thanks for the update! Can you please join https://www.w3.org/community/webassembly/ as well?

binji:
@RReverser For any significant changes, we're pretty strict. For small typo fixes, we usually don't require it. I guess the question is where this lands -- personally I'd say it could go either way.

RReverser:
@binji Could / should we include some sort of check as part of the CI? Is it even possible? It would certainly make it easier to communicate this to folks and get it fixed sooner if it was automated.

binji:
@RReverser yes, we have a check for this already in the spec repo, and we could add it here too. It requires a bit more than just joining the CG, however.

RReverser:
Hmm isn't that what you said to do in https://github.com/WebAssembly/website/pull/120#issuecomment-639182961?

binji:
Sorry, I mean the bot requires you to do more than just join the CG.

RReverser:
Ah ok. Do you think you could integrate it here if you already know how it works?

binji:
I didn't set it up originally, but now that it is already in the spec repo, I may be able to use the same thing here. I'll take a look. (here's the tool, in case you're curious https://github.com/w3c/ash-nazg/)

RReverser:
@kubo39 Ping. I'd love to merge this, but can you please confirm that you joined the CG?

kubo39:
@RReverser 
Sorry for late response. I'm unclear on whether I'm eligible to participate, I'm not employed by a W3C Member.
https://www.w3.org/participate/faq#individual

RReverser:
Sigh. Thanks for checking it out.

@binji I suppose I can go with "this is a small change" here, but do you think we could take the website from under these restrictions? Is it technically under W3C umbrella?

Ideally I'd want contributors to feel free to send patches without a complicated process, if possible.

binji:
@kubo39 You don't have to be employed by a W3C member, you can also join as an individual. But I understand if you don't want to do that.

@RReverser 
That reminds me: when we've discussed this before (w.r.t. projects like binaryen and wabt), we found that the main concern for w3c is the spec document, not the other artifacts (others please correct me if I'm misremembering). @ericprud is that correct?

RReverser:
I'm going to take a risk and proceed here. Thanks again @kubo39!

PR #167, opened 2020-06-15
Avoid repaints

surma:
I noticed that the hover effect on the side nav animates layout, which causes layout and paint on every frame. Here’s a quick fix to move that animation to the compositor.

### DIFF
--- a/css/custom.css
+++ b/css/custom.css
@@ -1564,13 +1564,11 @@ header {
     border-bottom: 1px solid #d2d2d2;
     padding-top: 5px;
     padding-bottom: 5px;
-    padding-right: 10px;
-    transition: .5s ease;
+    transition: transform .5s ease;
 }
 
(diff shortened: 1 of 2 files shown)


css/custom.css
```diff
@@ -1564,13 +1564,11 @@ header {
...
     padding-bottom: 5px;
-    padding-right: 10px;
-    transition: .5s ease;
+    transition: transform .5s ease;
 }
 
 .side-nav-item:hover {
-    padding-left: 10px;
-    padding-right: 0px;
+	transform: translateX(5px);
```
RReverser, line 1571: I know that file has mixed indentation already, but can you please fix this line back?

(I'll auto-format CSS at some point, but probably not in the scope of this PR)

surma: Oh shit, it looked fine on my screen. Stupid of me.

PR #162, opened 2020-05-27
Mark docs folder as autogenerated

RReverser:
This should help with reviews by collapsing any changes in `docs` by default.

PR #168, opened 2020-06-16
Switch to Github actions

RReverser:
I find `docs` folder in the repo:

 - Rather distracting for reviews (although the `.gitattributes` change in #162 helped a bit).
 - Unnecessarily complicating simple contributions for people who just edit Markdown using Github UI (most recent example being #166).

As such, I've decided to:
 - Create a `gh-pages` branch (done, left only contents of the current `docs` folder).
 - Switch from Travis to Github actions for building the website.
 - Leverage a Github action for building & deploying result to Github Pages automatically on any push to master.

Thoughts?


RReverser:
Note that I haven't removed the `docs` folder yet because I want to make sure that `gh-pages` actually gets built correctly first.

RReverser:
Deployment part doesn't seem to work correctly yet; feel free to review, but don't merge please. I'll take another look tomorrow.

docs/docs/js/index.html
```diff
@@ -983,22 +983,22 @@ <h2 id="sample-api-usage">Sample API Usage</h2>
...
 
-<pre><code class="language-lisp">(module
-    (import "js" "import1" (func $i1))
-    (import "js" "import2" (func $i2))
-    (func $main (call $i1))
-    (start $main)
-    (func (export "f") (call $i2))
-)
-</code></pre>
+<div class="language-lisp highlighter-rouge"><div class="highlight"><pre class="highlight"><code><span class="p">(</span><span class="nv">module</span>
```
binji, line 986: Can you update all of this in a separate PR?

RReverser: I guess yeah, this folder is supposed to go away altogether, this change was triggered by dependency update.

sbc100:
Great to see the switch to github actions!   We've had a lot of success with that over time.

In terms of moving the docs into a branch.. I actually find that model really confusing.  I guess its OK if you never have to deal with it locally. 

I really like being able to see the effect a PR has on the input and the output docs in the same PR and have the actions check that the input matches the output.    Is it just me that likes this?     Without this its hard to tell the effect on the generated docs that given change has, no?   

Id also like to be able to generate the docs locally and do a git diff to see the effect on the output locally,  having to switch branches with outstanding change is pretty tricky/confusing isn't it?

I agree there are downsides to casual contributors... ideally we could have an action that proposes the changes to output when one changes the input. 

Regardless, perhaps we could split this into 3 different PRs:

1. Switch to actions
2. Update the ruby gem deps
3. Propose moving the output docs into a different branch?

.github/workflows/main.yml
```diff
@@ -0,0 +1,32 @@
...
+    - name: Build
+      run: bundle exec jekyll build
+    - name: Deploy
+      # if: github.event_name == 'push'
+      uses: peaceiris/actions-gh-pages@v3
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        publish_dir: docs
+        user_name: 'github-actions[bot]'
+        user_email: 'github-actions[bot]@users.noreply.github.com'
```
sbc100, line 32: Does this end up making a second commit based on this commit?

sbc100:
For example, imagine we update a ruby dependency and that effects the generated site (but the input site is unchanged)... in this new proposed model how would we see the effect of the change in code review?

.github/workflows/main.yml
```diff
@@ -0,0 +1,32 @@
+name: CI
```
sbc100, line 1: Would it make sense to call this file CI.yaml to match the name?

RReverser:
> I really like being able to see the effect a PR has on the input and the output docs in the same PR and have the actions check that the input matches the output. Is it just me that likes this? Without this its hard to tell the effect on the generated docs that given change has, no?
> 
> Id also like to be able to generate the docs locally and do a git diff to see the effect on the output locally, having to switch branches with outstanding change is pretty tricky/confusing isn't it?

I'm not sure `git diff` on auto-generated HTML is very valuable for reviews (which is also the reason I've hid it in #162).

 - If the change is a minor update to text, most of the time it's straightforward enough and easier to inspect as a regular Markdown diff.
 - If it's related to styling, or just a big change to text that might affect layout etc., then HTML diff is not helpful either, and most of the time you (or at least me) want to build the website locally instead and visually inspect the modified pages in the browser and compare with the live `webassembly.org`.

Both of these workflows are still possible after this PR just like it was before.

> Regardless, perhaps we could split this into 3 different PRs:
> 
> 1. Switch to actions
> 2. Update the ruby gem deps
> 3. Propose moving the output docs into a different branch?

I could... It's just that, as mentioned in few other previous PRs, I've recently took over maintenance of `webassembly.org` and there's many more potentially vast changes in the pipeline.

My hope/understanding was that at least infrastructure updates would be mostly uncontroversial and up to me, so trying not to get too blocked on those and move faster :)

*(In these case, all of these steps are also dependent on each other, which would make reviews even slower due to waterfall-like nature.)*

sbc100:
Since I don't contribute to this repo myself, and you do, I feel like you get to have the final say here.

As I've said, I really don't like the "docs are hidden a different branch" approach of gh-pages, but if you value the advantaged you list I'm not going to strongly object.   

Ironically, I'm trying to do the exact opposite with emscripten right now and bring the generated docs  in-tree for easier one-step maintenance :)

RReverser:
> As I've said, I really don't like the "docs are hidden a different branch" approach of gh-pages

I think it's important to remember that this affects only live version of the website; locally workflow doesn't change in that you just start `bunde exec jekyll serve` or whatever and site is still getting built to a local folder, so everything else except diffing works just like before.

And then, even for diffing with "production" it only involves change in command to:

```shell
git diff HEAD:./docs gh-pages:./
```

so

> having to switch branches with outstanding change is pretty tricky/confusing isn't it?

doesn't really apply either.

Given this, I don't see any downsides TBH - contributors don't need to bother with build artifacts, and maintainers can still do everything they could before :)

.github/workflows/main.yml
RReverser: Maybe. FWIW this is the default template Github generates for "set up a workflow yourself":

image

So I assumed it's a convention for default filename / name?

RReverser:
> Ironically, I'm trying to do the exact opposite with emscripten right now and bring the generated docs in-tree for easier one-step maintenance :)

For reasons above I'd really love not to have to deal with build system for docs in Emscripten either, but I suppose I don't have a say there instead :P

sbc100:
> > Ironically, I'm trying to do the exact opposite with emscripten right now and bring the generated docs in-tree for easier one-step maintenance :)
> 
> For reasons above I'd really love not to have to deal with build system for docs in Emscripten either, but I suppose I don't have a say there instead :P

I think for drive-by edits, the way to fix that problem is to have "suggested edits" from the doc builder bot when the contributor does't want to install doc building toolchain.

sbc100:
We have the same problem with code contributions in emscripten and binaryen where its often a pain for drive-by code contributors to figure out how to update all the test expectations... "suggested edits" would be nice here too.

RReverser:
> I think for drive-by edits, the way to fix that problem is to have "suggested edits" from the doc builder bot when the contributor does't want to install doc building toolchain.

Yeah that fixes half of the problem - in fact, it's what I could do in this PR as well, there is Actions support for committing back to the same branch instead of `gh-pages`, but from review perspective I still find it easier to not deal with build artifacts either in Github reviews, or when pulling repo updates, or in `grep` / code search locally, or in any other places aside from actual live version.

In this regard I don't see any difference between generated HTML / CSS / JS and, say, binary object files, which we don't commit to repos either.

sbc100:
I don't mean to block this change by the way, feel free to ignore my opinion and commit this PR :)

RReverser:
Oh yeah I understand, I've just derailed conversation into general pros/cons in Emscripten as well. But this is not the best place for such conversation, of course :)

docs/docs/js/index.html
RReverser: I've left current contents of `docs` folder in `gh-pages` branch and switched production to use that as source of truth; now I've removed `docs` folder in this PR, so this change is no longer relevant.

Can you please take another look overall?

.github/workflows/main.yml
RReverser: Yeah commits in `gh-pages` branch are separate, since they're for autogenerated files not for authored ones.

You can see an example deployment commit here: https://github.com/WebAssembly/website/commit/fef1b8cf252dc1aa9bef396172bc5959b7a991da

RReverser:
Also cleaned up commit history from intermediate steps and temporary experiments.

RReverser:
@binji I'm assuming your approval still stands, but because I made few more small changes and cleanups, would prefer someone to make another glance before merging :)

RReverser:
I guess I'm going to go ahead and try this, since the live version already points to the new branch now.

PR #170, opened 2020-06-18
Add Node.js info on roadmap page

kripken:
The data has obvious overlap with Chrome (I guess like wasmtime and wasmer, and like Firefox and wasmtime are or will be) but it's not identical, and in particular the flags for enabling things are different.

The info is for current Node.js LTS (which is their recommended download).

Unfortunately this addition is beyond the visible size of the element, so one needs to scroll to see it, at least on my machine on Chrome and Firefox. Perhaps the layout could be changed somehow but that's definitely well beyond my abilities...


kripken:
About the test failure, I have no diff locally when I run `bundle exec jekyll build`. Is there a way to see what the diff on CI is, or another way to debug that?

RReverser:
Ugh, don't bother with the diff, it should go away with #168.

RReverser:
Re: addition itself, good point, I was assuming that most people will "translate" Chromium support to Node.js, but didn't take into account that they have different flags.

I'll review properly tomorrow when back at my computer.

RReverser:
> Ugh, don't bother with the diff, it should go away with #168.

Merged that one and rebased this PR.

features.json
```diff
@@ -124,6 +124,22 @@
 				"simd": "--enable-simd",
 				"threads": "--enable-threads"
 			}
+		},
+		"Node.js": {
+			"url": "https://nodejs.org/",
+			"logo": "/images/nodejs.svg",
+			"version": "12.18.1",
```
RReverser, line 131: I can see benefits of using LTS version for some customers, but TBH I'd just use the "current" stable one (14.4.0) for the feature table instead.

Otherwise, for example, Firefox also has an ESR (Extended Support Release) version that we could use, but I'm not sure it's interesting to *most* consumers.

RReverser:
> Unfortunately this addition is beyond the visible size of the element, so one needs to scroll to see it, at least on my machine on Chrome and Firefox. Perhaps the layout could be changed somehow but that's definitely well beyond my abilities...

Yeah the current number of columns _just_ fits, scrolling is pretty much unavoidable.

As we add more runtimes, we might add filtering via checkboxes for different categories of engines (sort of like caniuse does), but for now horizontal scroll should be fine.

features.json
kripken: Unlike Firefox though, Node.js does default to the LTS on its download page. I assume that means most users are on that, but I'm not sure how to check?

(I'm also assuming using the version most users are on is best for the table, is that a good assumption?)

RReverser: > (I'm also assuming using the version most users are on is best for the table, is that a good assumption?)

Yeah, that's perfectly reasonable IMO.

> Unlike Firefox though, Node.js does default to the LTS on its download page. I assume that means most users are on that, but I'm not sure how to check?

Hmm, that's a good point. TBH I never visited that page directly, and the home page (which also contains direct download links) shows both as equal: https://nodejs.org/en/

kripken:
About the scrolling issue, we currently have long names in the first column, like `JS BigInt to Wasm i64 integration`, on a single line - maybe those could be split to two lines?

(It is kind of nice how caniuse.com shows ~15 VMs without scrolling, I think.)

features.json
RReverser: > I assume that means most users are on that, but I'm not sure how to check?

Maybe @MylesBorins can help?

RReverser:
Sure, but that would be a temporary solution anyway - e.g. I was thinking about adding Wasm3, and another person on Twitter mentioned WAVM as potentially popular engine to add...

We could do that, just thinking about how long it will hold / whether it's worth the time. FWIW initial list didn't fit without scrolling either, which is why I had to rename "JavaScript BigInt to WebAssembly i64 integration" to a shorter version you quoted, but, as we can see, that was already rather temporary :)

features.json
RReverser: Meanwhile, do you know what the differences are in terms of feature support between those versions?

kripken: Meanwhile the best data I can find shows LTS is 55% of users, so it is most popular, but by less than I would have guessed, https://nodejs.org/en/user-survey-report/#version . Latest is 43%.

But the Node.js download page, and top links for "which node.js version?" do point to LTS (e.g. https://tamalweb.com/which-nodejs-version)

Checking the differences as you asked, it turns out there is no difference in features actually (or flag names - only difference I see is the new `--wasm-staging` option was added). Given no differences, showing the older one seems more general?

RReverser: Thanks for the analysis! It seems quite close to 50/50, so even though LTS is slightly more popular according to the survey, I would be still vary of preferring one over another under different circumstances. However...

> Checking the differences as you asked, it turns out there is no difference in features actually (or flag names - only difference I see is the new `--wasm-staging` option was added). Given no differences, showing the older one seems more general?

is exactly the reason I've asked about differences, and I'm glad to hear there isn't any. If so, yeah, I agree we should include LTS / older version here as it's more general.

We'll probably have to revisit this question in the future when stable Node.js gains a new Wasm feature before LTS does, but for now this should be good to go.

RReverser, line 131: Let's omit the non-important patch version, like in other engines/browsers. It shouldn't ever affect supported feature set, and only makes output a bit more verbose.

```suggestion
			"version": "12.18",
```

features.json
```diff
@@ -124,6 +124,22 @@
 				"simd": "--enable-simd",
 				"threads": "--enable-threads"
 			}
+		},
+		"Node.js": {
+			"url": "https://nodejs.org/",
+			"logo": "/images/nodejs.svg",
```
RReverser, line 130: Looking at local output, this logo seems out of line compared to others, and details are not very readable at this size anyway:
image

Could we use the icon version instead? It seems to be allowed as a "primary icon + pattern" in [official visual guidelines](https://nodejs.org/static/documents/foundation-visual-guidelines.pdf) and would look better at this size.

I don't know where to get SVG version for it, but then, I'm not sure where you got this one either TBH 😀

kripken: I got it from wikipedia, https://en.wikipedia.org/wiki/Node.js#/media/File:Node.js_logo.svg

I don't see an icon version there or on https://nodejs.org/en/about/resources/

Doing some more searching, I can't find a download for the "primary icon + pattern" anywhere... so I used Inkscape to crop the existing logo...

RReverser: I found one in their page source but it seems to be a sprite image and would need to be cropped or manually extracted too.

**UPD**: actually, looks like that sprite is built from separate sources, and the one we want is here: https://github.com/nodejs/nodejs.org/blob/master/static/images/logo-hexagon.svg

RReverser: I'm fine with keeping either of those, just need to check if SVGs are optimised as part of the build or we need to compress the current version with SVGO...

RReverser: Looks like we don't; can you please either compress this SVG or add the one from Node repo (which looks already pre-compressed)?

kripken: Sure, I used the one from that link (since I have no idea how to compress one myself...)

RReverser: Oh, just in case you ever need that in the future - as I mentioned above, there is SVGO (probably the most popular SVG optimiser out there), and you can use it either via CLI, or, for example, via this Web UI by one of our fellow Web DevRels :) https://jakearchibald.github.io/svgomg/

RReverser: Thanks for updating!

RReverser:
Ok I think we'll want to fiddle with the styling of the horizontal scroll, but for now should be fine to merge as-is.

Thanks @kripken!

PR #174, opened 2020-06-24
Update Firefox and Chrome features support data

chicoxyzzy:
Firefox links
BigInt https://bugzilla.mozilla.org/show_bug.cgi?id=1631702#c1
bulkmem https://bugzilla.mozilla.org/show_bug.cgi?id=1528294#c0
multi-value https://bugzilla.mozilla.org/show_bug.cgi?id=1628321#c4
simd https://bugzilla.mozilla.org/show_bug.cgi?id=1625130

Chrome links
BigInt https://bugs.chromium.org/p/v8/issues/detail?id=7741#c21

### DIFF
--- a/features.json
+++ b/features.json
@@ -85,7 +85,7 @@
 				"referenceTypes": "Nightly",
 				"saturatedFloatToInt": true,
 				"signExtensions": true,
-				"simd": "x86 only",
+				"simd": "Nightly, x86 only",
 				"threads": "Nightly"
 			}
 		},
--- a/features.json
+++ b/features.json
@@ -62,7 +62,7 @@
 			"logo": "/images/chrome.svg",
 			"version": "83",
 			"features": {
-				"bigInt": "#enable-experimental-webassembly-features",
+				"bigInt": "Canary 85",
 				"bulkMemory": true,
 				"multiValue": "Canary",
 				"mutableGlobals": true,
--- a/features.json
+++ b/features.json
@@ -64,7 +64,7 @@
 			"features": {
 				"bigInt": "#enable-experimental-webassembly-features",
 				"bulkMemory": true,
-				"multiValue": "#enable-experimental-webassembly-features",
+				"multiValue": "Canary",
 				"mutableGlobals": true,
 				"referenceTypes": "#enable-experimental-webassembly-features",
 				"saturatedFloatToInt": true,
@@ -78,13 +78,14 @@
 			"logo": "/images/firefox.svg",
 			"version": "77",
 			"features": {
-				"bigInt": "Nightly",
-				"bulkMemory": "javascript.options.shared_memory",
-				"multiValue": "Nightly",
+				"bigInt": "Beta 78",
+				"bulkMemory": "Nightly",
+				"multiValue": "Beta 78",
 				"mutableGlobals": true,
 				"referenceTypes": "Nightly",
 				"saturatedFloatToInt": true,
 				"signExtensions": true,
+				"simd": "x86 only",
 				"threads": "Nightly"
 			}
 		},
--- a/features.json
+++ b/features.json
@@ -85,7 +85,7 @@
 				"referenceTypes": "Nightly",
 				"saturatedFloatToInt": true,
 				"signExtensions": true,
-				"simd": "Nightly, x86 only",
+				"simd": "Nightly, x86 and x86-64 only",
 				"threads": "Nightly"
 			}
 		},
--- a/features.json
+++ b/features.json
@@ -62,9 +62,9 @@
 			"logo": "/images/chrome.svg",
 			"version": "83",
 			"features": {
-				"bigInt": "Canary 85",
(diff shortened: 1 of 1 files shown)


RReverser:
Thanks for the PR. However, this table is currently designed to show a single version - latest stable - of each browser, or a flag for that stable version if feature is not finalised yet, or, if none of these apply, we can put "Canary" or "Nightly".

I don't really want to disrupt that order, as in most cases people want to target stable version of the browser, and fallback to nightly builds only if experimentation is otherwise impossible. As far as I understand, the status of these features on stable didn't change?

We might make the table to show the "minimum supported browser version" in the future, but that requires much more digging into data and increases time required for its maintenance compared to just "status on stable".

chicoxyzzy:
I reverted Chrome changes, will provide clarifications regarding Firefox changes

features.json
```diff
@@ -78,13 +78,14 @@
 			"logo": "/images/firefox.svg",
 			"version": "77",
 			"features": {
-				"bigInt": "Nightly",
-				"bulkMemory": "javascript.options.shared_memory",
-				"multiValue": "Nightly",
+				"bigInt": "Beta 78",
```
chicoxyzzy, line 81: This is enabled by default since Firefox 78, so Nightly is not correct status

features.json
```diff
@@ -78,13 +78,14 @@
...
+				"bulkMemory": "Nightly",
```
chicoxyzzy, line 82: This is available in Nightly without any flags

features.json
```diff
@@ -78,13 +78,14 @@
...
+				"multiValue": "Beta 78",
```
chicoxyzzy, line 83: This is enabled by default since Firefox 78

features.json
chicoxyzzy: it seems that it's because `javascript.options.shared_memory` is enabled in Nightly. I'll revert this also

features.json
RReverser: Is it now available via flags on stable as well perhaps?

chicoxyzzy: I can't find any related wasm flags on stable

image

Screen from Firefox 78 with all flags with default values

image

chicoxyzzy: comment about BigInt support in Beta https://bugzilla.mozilla.org/show_bug.cgi?id=1631702#c1
patch for multi-value support in Beta https://bugzilla.mozilla.org/show_bug.cgi?id=1628321#c3

RReverser: Sounds good, thanks for checking!

features.json
```diff
@@ -78,13 +78,14 @@
...
-				"bigInt": "Nightly",
+				"bigInt": "Beta 78",
 				"bulkMemory": "javascript.options.shared_memory",
-				"multiValue": "Nightly",
+				"multiValue": "Beta 78",
 				"mutableGlobals": true,
 				"referenceTypes": "Nightly",
 				"saturatedFloatToInt": true,
 				"signExtensions": true,
+				"simd": "Nightly, x86 only",
```
RReverser, line 88: Just in case - is this w/o flags?

chicoxyzzy: there is `javascript.options.wasm_simd` flag (in the Nightly version only) which is enabled by default

RReverser: Sigh, okay. Another clarification - where is information about "x86" coming from? Does it include x86-64? It's not very obvious IMO if so.

chicoxyzzy: Judging by dependat bugs of [an umbrella SIMD bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1625130), x86/x64 support is ready, arm32 and arm64 is work in progress. AFAIK, arm support landed in Cranelift couple weeks ago.

Do you want me to replace x86 by x86/x64? I meant the [x86 family](https://en.wikipedia.org/wiki/X86) which includes 16-bit, 32-bit and 64-bit versions, but I agree that this is not obvious. I'd be happy to make it more clean if necessary.

RReverser: Well Firefox likely doesn't run on 16-bit platforms, and 64-bit is traditionally encoded as x86-64 to explicitly distinguish from regular x86. You can see even in the link you provided `x86` only includes 16-bit and 32-bit versions, with `x86-64` being a separate group of rows.

Let's do `x86 and x86-64 only`.

lars-t-hansen: > AFAIK, arm support landed in Cranelift couple weeks ago.

This is not correct.  Some arm64 work is ongoing in Cranelift, which has a largely functioning arm64 backend.  No arm(32) work is happening in Cranelift to my knowledge.

Even though the matter has been resolved to my satisfaction, for the sake of record: IMO we should always record x86 and x64 aka x86-64 separately, as they are separate if closely related back-ends.

lars-t-hansen:
FWIW, bulk memory and reference types will ship in FF79 and will be in FF79 beta when that ships next week.

RReverser:
Thanks @chicoxyzzy!

PR #175, opened 2020-06-24
Group features by status

RReverser:
This is the first usage of `phase` field.

Note that precise value doesn't matter (which decreases the pressure put on us in keeping those in sync), but "phase 4" vs "the rest" seems like a useful split for developers.

## Before
image

## After
image


RReverser:
OTOH, our data currently merges "phase 5" and "phase 4" into a single group. Do you think it's useful to separate those out, even those both are kinda complete?

binji:
Generally, there should be little difference between phase 4 and phase 5, so I think it's OK to group them together.

RReverser:
Ok, I'll go ahead with this then.

PR #177, opened 2020-07-02
Firefox 78 updates

chicoxyzzy:
@lars-t-hansen [said](https://github.com/WebAssembly/website/pull/174#issuecomment-649455607) that bulkmem and multi-value will land in FF79, but I see those flags enabled by default in FF 78 stable 🤔

### DIFF
--- a/features.json
+++ b/features.json
@@ -76,16 +76,16 @@
 		"Firefox": {
 			"url": "https://www.mozilla.org/firefox/",
 			"logo": "/images/firefox.svg",
-			"version": "77",
+			"version": "78",
 			"features": {
-				"bigInt": "Beta 78",
-				"bulkMemory": "javascript.options.shared_memory",
-				"multiValue": "Beta 78",
+				"bigInt": true,
+				"bulkMemory": true,
+				"multiValue": true,
 				"mutableGlobals": true,
 				"referenceTypes": "Nightly",
 				"saturatedFloatToInt": true,
 				"signExtensions": true,
-				"simd": "Nightly, x86 and x86-64 only",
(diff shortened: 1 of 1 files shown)


chicoxyzzy:
cc @RReverser for review

lars-t-hansen:
Hm, multi-value and bigint<->int64 are supposed to be in FF78 but reftypes+bulk memory were supposed to wait for FF79.  But I see reftypes are indeed enabled in FF78.0.1, this is a little surprising.  @eqrion, did something get uplifted?

eqrion:
@chicoxyzzy Are you actually seeing reference-types usable in Fx78 or just that the pref is enabled? I see the pref is enabled, but it should be ignored outside of nightly builds (which is confusing and will be fixed).

I ran a couple tests and wasn't able to use any reference-types features. e.g, try out [1] and note reference-types.

Bulk-memory, however, does appear to be enabled in Fx78 by accident. I'm not sure what we will do about that yet, no actual changes were made to the feature since Fx78; we were just waiting for the community-group to advance to phase 4.

[1] https://wasm-feature-detect.surma.technology/

RReverser:
@eqrion FWIW https://webassembly.org/roadmap/ (from this repo) uses same tests as the page you linked :)

But yeah, IIUC, the question was about bulk memory which suddenly appeared as enabled.

chicoxyzzy:
@eqrion yes, reference types are non-usable, but the pref is turned on

eqrion:
Okay, that makes sense. There was a miscommunication then, and we were trying to figure out what to do about reference-types shipping earlier which would have been more difficult to deal with.

PR #178, opened 2020-07-02
Restructure design docs

RReverser:
As per https://github.com/WebAssembly/design/issues/1349, we don't want to keep most of the design docs at the top level as they're out of date.

Rather than getting rid of all of them at once, I've set up redirects for design docs that have explicit up-to-date versions (such as Binary Encoding, Text Format, Semantics, etc.), as well as for couple of docs that are no longer of interest - for those I'm redirecting to Markdown files in the original repo just so that URLs keep working.

For a few docs still under the question of whether they should be pointing to better resources or updated themselves, I'm currently keeping them in the sidebar, but merging Docs & FAQ sections together to reduce the clutter.


RReverser:
## Before

image

## After

image

RReverser:
@binji @tschneidereit Ping?

RReverser:
I guess I'll just go for it.

PR #181, opened 2020-07-16
Update Safari feature support

tadeuzagallo:
Safari/JSC supports reference types behind the `JSC_useWebAssemblyReferences` runtime flag.

### DIFF
--- a/features.json
+++ b/features.json
@@ -95,7 +95,8 @@
 			"version": "13.1",
 			"features": {
 				"multiValue": true,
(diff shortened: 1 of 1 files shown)


RReverser:
Awesome, thanks! Just to be clear - is this flag supported under Safari 13.1? If not, might be worth extending test to specify version (like we did in couple of places for Nightly or Beta versions of browsers).

tadeuzagallo:
Yes, it's supported in 13.1.

PR #183, opened 2020-07-28
Update Firefox version in the feature support table

chicoxyzzy:
https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/79#WebAssembly

PR #184, opened 2020-08-06
Update features.json for Firefox 79

fitzgen:
@RReverser, please take a look, thanks!

### DIFF
--- a/features.json
+++ b/features.json
@@ -76,17 +76,17 @@
 		"Firefox": {
 			"url": "https://www.mozilla.org/firefox/",
 			"logo": "/images/firefox.svg",
-			"version": "78",
+			"version": "79",
 			"features": {
 				"bigInt": true,
 				"bulkMemory": true,
 				"multiValue": true,
 				"mutableGlobals": true,
-				"referenceTypes": "Nightly",
+				"referenceTypes": true,
 				"saturatedFloatToInt": true,
 				"signExtensions": true,
-				"simd": "Beta 79, x86 and x86-64 only",
-				"threads": "Beta 79"
+				"simd": "x86 and x86-64 only",
+				"threads": true
 			}
 		},
 		"Safari": {


features.json
```diff
@@ -82,11 +82,11 @@
 				"bulkMemory": true,
 				"multiValue": true,
 				"mutableGlobals": true,
-				"referenceTypes": "Nightly",
+				"referenceTypes": true,
 				"saturatedFloatToInt": true,
 				"signExtensions": true,
-				"simd": "Beta 79, x86 and x86-64 only",
-				"threads": "Beta 79"
+				"simd": "x86 and x86-64 only",
```
kripken, line 88: Oh nice, is SIMD enabled by default on x86/x86_64 now?

RReverser:
Just to double-check - it went stable, right? You need to update the `"version"` field too, but otherwise LGTM!

features.json
```diff
@@ -82,11 +82,11 @@
 				"bulkMemory": true,
```
kripken, line 82: a few lines above this one there is `"version": "78",` which I think should be updated to 79?

RReverser: Yeah, see my comment above, now resolved :)

features.json
RReverser: Yeah it was on Beta too, so looks that way!

fitzgen:
Woops, I missed the version field. Fixed now!

@kripken 
> Oh nice, is SIMD enabled by default on x86/x86_64 now?

I thought so, but it looks like it actually didn't ride the train to stable. I've force pushed to update that

fitzgen:
Aaaaaand it looks like this PR landed 34 seconds too soon :sweat_smile:

features.json
kripken: Odd, the Firefox 79 release notes don't mention it,

https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/79

Perhaps an oversight in the notes? Or maybe it wasn't flipped on?

RReverser:
Arghhh sorry I was excited :D

fitzgen:
Should I make a new PR?

RReverser:
I can update manually, but what should it say? In Beta they were enabled by default (if I remember correctly), has that changed in stable to a flag or something?

RReverser:
I don't see a flag in `about:config` either...

fitzgen:
I think the "Beta 79" prefix that it originally had can stay.



> I don't see a flag in `about:config` either...

I am just basing this off of (a) the lack of SIMD mention in https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/79#WebAssembly and (b) when I open the roadmap page in Firefox 79 stable it has an X for SIMD in the "your browser" column.

It is _possible_ that it did ride to stable, happened not to get mentioned in the release notes, and the wasm feature detection library has a bug, but seems fairly unlikely...

RReverser:
> I think the "Beta 79" prefix that it originally had can stay.

I see, I guess it's "Beta 80" then now?

RReverser:
> It is possible that it did ride to stable, happened not to get mentioned in the release notes, and the wasm feature detection library has a bug, but seems fairly unlikely...

Yeah I think we saw it passing on Beta version before, so unlikely a bug. Let me try with Beta 80 to be sure.

fitzgen:
Well, teh support does exist in Beta 79 too, assuming it was correct before. Kinda confusing that it is in beta 79 but not stable 79 :-/

RReverser:
> Kinda confusing that it is in beta 79 but not stable 79 :-/

Indeed, I've just assumed that they're auto-propagated to stable release and didn't bother to double-check like @kripken did. Should be more careful next time :)

> Let me try with Beta 80 to be sure.

Hmm, just tried and it shows "X" there as well...

RReverser:
I've tried few more channels and for now only Nightly seems to pass the test. Hence, I'll put "Nightly" prefix as the most up-to-date information, until something changes and a stable-ish channel supports SIMD again.

RReverser:
Oh wow I missed https://github.com/WebAssembly/website/pull/183 which was sent while I was taking time off, where @chicoxyzzy seemed to do the same work and ran into the same conclusion.

features.json
lars-t-hansen: SIMD is Nightly-only.  There's work ongoing to let it ride the train to Release, but preffed off (since Phase 3).

PR #186, opened 2020-09-15
Update Wasm supported features in Wasmer

syrusakbary:
We just published a new version of Wasmer supporting multivalue and bulk memory proposals 😊

### DIFF
--- a/features.json
+++ b/features.json
@@ -117,9 +117,11 @@
 		"Wasmer": {
 			"url": "https://wasmer.io/",
 			"logo": "/images/wasmer.png",
-			"version": "0.17",
+			"version": "1.0",
 			"features": {
 				"bigInt": true,
+				"bulkMemory": true,
+				"multiValue": true,
 				"mutableGlobals": true,
 				"saturatedFloatToInt": true,
 				"signExtensions": true,


RReverser:
1. It looks like 1.0 is in alpha for now? We use stable versions of each engine in the table.
2. Are these features enabled by default or require a flag?

syrusakbary:
> It looks like 1.0 is in alpha for now? We use stable versions of each engine in the table.
That's right! We are actually already shipping the "alpha" version when someone runs [Wasmer install command](https://github.com/wasmerio/wasmer-install#install-latest-version), so that's why I thought it could be useful to reflect it here :)


> Are these features enabled by default or require a flag?
multivalue and bulk memory are now [enabled by default](https://github.com/wasmerio/wasmer/blob/master/lib/wasmer-types/src/features.rs#L30-L33) (that means: no flag should be required)

RReverser:
> That's right! We are actually already using the "alpha" version when someone [installs Wasmer](https://github.com/wasmerio/wasmer-install#install-latest-version), so that's why I thought it could be useful to reflect it here :)

Hmm, seems dangerous, but I guess if you count it as stable enough to be default, then makes sense.

PR #215, opened 2020-10-14
Fix 404 link to exceptions proposal in features.json

MaxDesiatov:


RReverser:
Ah looks like upstream made some reorg in the repo. Thanks!

PR #221, opened 2020-11-02
Represent non-applicable roadmap features with 'N/A'. Fixes #171

### LINKED ISSUE #171: Roadmap checkmarks are ambiguous
"JS BigInt to Wasm i64 integration" has a checkmark for wasmtime and wasmer, which I guess means "all necessary support is present", that is, they have no JS, so no support is needed, so it's implemented there "in the empty sense". Is that right?

This does make sense in that someone looking at the *row* for "JS BigInt to Wasm i64 integration" probably wants to see a checkmark there, if they are curious if they can ship with that feature.

However, it is ambiguous and it confused me. In particular, when looking at the *column* for a JS-less VM and wondering what has been implemented there, it gives the wrong answer.

Perhaps there could be a non-ambiguous symbol for "irrelevant" or "unnecessary"? I don't have a great idea, but maybe a green checkmark for "implemented" and a gray checkmark for "irrelevant"? Or if color isn't enough, a circle for "irrelevant"? (which is clearly neither a checkmark, an X, or an hourglass)

kripken:
After looking at the various unicode options, I think minus ➖ makes the most sense.

Fixes #171

Sorry I couldn't test this, when I try to build I get this:
```
 Incremental build: disabled. Enable with --incremental
      Generating... 
  Liquid Exception: undefined method `includes_load_paths' for #<Jekyll::Site:0x000055c2eed09e70> in _layouts/getting-started.html
jekyll 3.1.6 | Error:  undefined method `includes_load_paths' for #<Jekyll::Site:0x000055c2eed09e70>
```

I can try again from my work machine tomorrow.

### DIFF
--- a/features.json
+++ b/features.json
@@ -104,7 +104,7 @@
 			"logo": "/images/bca.png",
 			"version": "0.20",
 			"features": {
-				"bigInt": true,
+				"bigInt": null,
 				"bulkMemory": true,
 				"multiValue": true,
 				"mutableGlobals": true,
@@ -119,7 +119,7 @@
 			"logo": "/images/wasmer.png",
 			"version": "1.0",
 			"features": {
-				"bigInt": true,
+				"bigInt": null,
 				"bulkMemory": true,
 				"multiValue": true,
 				"mutableGlobals": true,
--- a/roadmap.md
+++ b/roadmap.md
@@ -100,7 +100,7 @@ After the initial release, WebAssembly has been gaining new features through the
                 return h('td', { title: support, tabIndex: 0 }, ['⏳']);
               }
               if (support === null) {
-                return h('td', { title: support, tabIndex: 0 }, ['➖']);
+                return h('td', { title: support, tabIndex: 0 }, ['ⁿ/ₐ']);
               }
               return h('td', {}, [support ? '✔️' : '❌']);
             })
--- a/roadmap.md
+++ b/roadmap.md
@@ -99,6 +99,9 @@ After the initial release, WebAssembly has been gaining new features through the
               if (typeof support === 'string') {
                 return h('td', { title: support, tabIndex: 0 }, ['⏳']);
               }
+              if (support === null) {
+                return h('td', { title: support, tabIndex: 0 }, ['➖']);
+              }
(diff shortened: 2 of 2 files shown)


RReverser:
Thanks. Looking at it, I'm a bit worried that `➖` might be confusing alongside `❌`. I agree that there doesn't seem to be a better single Unicode char, but how about something like `ⁿ/ₐ`? It's technically not an emoji nor a single char, but seems to look quite naturally:

image

kripken:
Sounds good, done.

RReverser:
Thanks! One more thing - for now you marked only JS-BigInt integration, but I suppose we should do the same for threads too?

It seems similar because, just like in BigInt case, the Wasm part would compile, but the feature is likely meaningless without the host-side integration, and WASI doesn't have support for creating threads yet.

kripken:
I'm not sure about threads. While it's true that WASI doesn't yet have threads, that's not the same as "not applicable". BigInt is really not applicable, but threads is applicable, just waiting for some work to be done (including spec work, in this case). Also, wasmer has support behind a flag (according to the existing data). So I'd rather not change anything on threads in this PR.

RReverser:
> Also, wasmer has support behind a flag

It has support for atomics (which is why I added it), but not threads themselves. I guess it's true about "work to be done". I'm a bit on the fence, but makes sense to merge this as-is for now.

RReverser:
It's out, but looks like it's marking it with a tooltip... I'll do a quick fix.

RReverser:
(I probably shouldn't have been merging this on weekend night...)

RReverser:
Oh, right, it's because the `title` attribute was copied as-is and is explicitly passing `null`, heh.

RReverser:
Ok, it's fixed now.

kripken:
I probably shouldn't have written this on a weekend night either :wink: thanks for merging and fixing!

RReverser:
> I probably shouldn't have written this on a weekend night either 😉

Totally :) Wish Github had a way to mark work-related repos separately from "fun" ones and mute them for the weekend.

PR #225, opened 2021-01-08
Wasmtime 0.22

fitzgen:
Can you take a look at this @RReverser? Thanks!

### DIFF
--- a/features.json
+++ b/features.json
@@ -15,6 +15,11 @@
 			"url": "https://github.com/WebAssembly/exception-handling/blob/master/proposals/exception-handling/Exceptions.md",
 			"phase": 2
 		},
+		"moduleLinking": {
+			"description": "Module Linking",
+			"url": "https://github.com/WebAssembly/module-linking/blob/master/proposals/module-linking/Explainer.md",
+			"phase": 1
+		},
 		"multiValue": {
 			"description": "Multi-value",
 			"url": "https://github.com/WebAssembly/spec/blob/master/proposals/multi-value/Overview.md",
--- a/features.json
+++ b/features.json
@@ -107,16 +107,18 @@
 		"Wasmtime": {
 			"url": "https://wasmtime.dev/",
 			"logo": "/images/bca.png",
-			"version": "0.20",
+			"version": "0.22",
 			"features": {
 				"bigInt": null,
 				"bulkMemory": true,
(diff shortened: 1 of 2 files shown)


roadmap.md
```diff
@@ -79,7 +79,11 @@ After the initial release, WebAssembly has been gaining new features through the
       );
       for (let { name, description, url, phase } of features) {
         let supportHTML = h('td');
-        wasmFeatureDetect[name]()
+        Promise.resolve()
+          // Make sure to call `wasmFeatureDetect` inside a promise-chained
+          // function so that we can still render table rows for features that
+          // we don't have a detector yet.
+          .then(() => wasmFeatureDetect[name]())
```
RReverser, line 86: Makes sense. Do you want to submit a PR to add a detector for module linking though?

RReverser (approved):
Thanks!

roadmap.md
fitzgen: Already filed https://github.com/GoogleChromeLabs/wasm-feature-detect/issues/38 -- it is a little more complicated because WABT doesn't support module linking yet.

fitzgen:
Thanks!

features.json
```diff
@@ -102,16 +107,18 @@
...
 				"bulkMemory": true,
+				"moduleLinking": "--enable-module-linking",
 				"multiValue": true,
 				"mutableGlobals": true,
 				"referenceTypes": true,
 				"saturatedFloatToInt": true,
 				"signExtensions": true,
-				"simd": "--enable-simd"
+				"simd": "--enable-simd",
+				"threads": "--enable-threads"
```
RReverser, line 121: @fitzgen I just remembered why I didn't enable this in the first place. Is the issue I reported at https://github.com/bytecodealliance/wasmtime/issues/1658 resolved?

fitzgen: We don't have shared memories yet, but we have partial support for atomics instructions. I was under the impression that the flags represented ongoing, work-in-progress support, which is why I added it. Maybe I misunderstood, and the flags should only be added when it is "done" but not on by default yet?

RReverser: > and the flags should only be added when it is "done" but not on by default yet?

Yeah, basically that.

RReverser: Sorry for the confusion!

PR #233, opened 2021-04-07
Update flag for Firefox SIMD

RReverser:
It seems to be possible to enable it on stable too now.

@lars-t-hansen can you please confirm if it's still limited to x86 / x64?

### DIFF
--- a/features.json
+++ b/features.json
@@ -90,7 +90,7 @@
 				"referenceTypes": true,
 				"saturatedFloatToInt": true,
 				"signExtensions": true,
-				"simd": "Nightly, x86 and x86-64 only",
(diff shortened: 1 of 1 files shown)


RReverser:
Merging as-is for now, we can fixup later.

lars-t-hansen:
@RReverser, SIMD is currently only x64/x86, enabled by default in beta/release from FF89 (firm).

ARM64 support in Nightly soon, enabled by default in beta/release from FF90 (likely).

RReverser:
@lars-t-hansen Sounds good, thanks! In that case we'll update flags again once FF89 is out.

Issue #231, opened 2021-03-20
Update Microsoft Edge logo on home page


alvinashcraft:
The Edge logo on the home page is the old "E" logo used by the non-Chromium based Edge. This should be updated to use the new logo. [https://en.wikipedia.org/wiki/Microsoft_Edge#/media/File:Microsoft_Edge_logo_(2019).svg](url)

sidverma32:
Hi @alvinashcraft  , Shall i pick this up ? As this will be my first contribution to the repo.

alvinashcraft:
Yes, this one sounds like a great first contribution, if you are interested.

chicoxyzzy:
Fixed in https://github.com/WebAssembly/website/pull/283/commits/fb84f2abd4525371ff70c7609d8bff4fb358eaf1#diff-f12839dc24ee04589e18aec3114fc2c2727da5b3e219a8c42cac7cbce18d4200

PR #239, opened 2021-05-11
Updated Microsoft Edge logo on home page

sidverma32:
### Problem
The Edge logo on the home page is the old "E" logo used by the non-Chromium based Edge.

### ChangeLog

Updated the Edge browser logo to a new one.


RReverser:
> The Edge logo on the home page is the old "E" logo used by the non-Chromium based Edge.

Hm, thinking about this a bit more... the point of that row was to say that 4 major engines (including the non-Chrome Edge) has shipped WebAssembly support.

If we switch to the Chromium-based Edge logo, we might as well remove its mention altogether, because in terms of engine it's the same one as Chrome...

sidverma32:
@RReverser Thanks for describing the problem statement. But i am little less familiar with the repo., if you can give me a heads-up on the pointer which you said  like from where to begin.

andylizi:
This change was merged in #158, which I included as part of another project because I hadn't noticed you've already submitted this PR. I apologize for wasting your time:pensive:.

#231 can also be closed now.

chicoxyzzy:
Fixed in #283

PR #241, opened 2021-06-02
Update Wasmer features

syrusakbary:
Wasmer 2.0 (which will be released in 15 hours), fully supports the latest phase 4 WebAssembly proposals (SIMD and Reference Types)


syrusakbary:
Just released the `2.0.0-rc1` which is now the default option when installing Wasmer from the shell! 🎉

PR #247, opened 2021-07-25
Update tinygo getting started link

iwgx:
The current link is not worked anymore, so I'm update it to the appropriate one. I hope this is right!

Screenshot 2021-07-25 143458


RReverser (approved):
Thanks, good catch!

PR #250, opened 2021-09-20
Update Safari Feature Support for 15.0

CryZe:
Safari 15 just released with `Bulk memory operations`, `Reference types` and `Non-trapping float-to-int conversions` now being activated by default.

### DIFF
--- a/features.json
+++ b/features.json
@@ -106,14 +106,14 @@
 		"Safari": {
 			"url": "https://www.apple.com/safari/",
 			"logo": "/images/safari_48x48.png",
-			"version": "14.1",
+			"version": "15.0",
 			"features": {
 				"bigInt": true,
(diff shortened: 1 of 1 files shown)


RReverser:
Nice, thanks! I vaguely remember someone saying COOP/COEP might come in Safari 15 too, is not enabled by default yet?

CryZe:
I'm not familiar with what those do, but a quick google seems to reveal it's related to SharedArrayBuffer and thus maybe the `Threads and atomics`? That feature is still shown as unavailable when I visit the site and check the `Your browser` column.

RReverser:
Yeah, they're related to enabling SAB & threads. Oh well, thanks!

RReverser (approved):

chicoxyzzy:
@RReverser COEP/COOP are only available in Safari TP under flag. I can be wrong, but as far as I know, implementation is not complete yet.

PR #257, opened 2021-12-14
Update Safari version to 15.2

chicoxyzzy:
Post from WebKit blog: https://webkit.org/blog/12140/new-webkit-features-in-safari-15-2/

Bug on WebKit Bugzilla: https://bugs.webkit.org/show_bug.cgi?id=229681

### DIFF
--- a/features.json
+++ b/features.json
@@ -117,10 +117,11 @@
 		"Safari": {
 			"url": "https://www.apple.com/safari/",
 			"logo": "/images/safari_48x48.png",
-			"version": "15.0",
+			"version": "15.2",
 			"features": {
 				"bigInt": true,
 				"bulkMemory": true,
+				"exceptions": true,
 				"multiValue": true,
 				"mutableGlobals": true,
 				"referenceTypes": true,


RReverser:
Looks like we should be able to flip `threads` to `true` as well.

Is this release rolled out on stable iOS & macOS now?

chicoxyzzy:
> Is this release rolled out on stable iOS & macOS now?

Yep https://developer.apple.com/documentation/safari-release-notes/safari-15_2-release-notes

> Looks like we should be able to flip threads to true as well.

I think it's still under flag.

RReverser:
> I think it's still under flag.

If I'm reading this right:

> Safari 15.2 adds support for Cross-Origin-Opener-Policy (COOP) and Cross-Origin-Embedder-Policy (COEP) HTTP response headers. Sites can adopt these headers to opt into process isolation and be better protected. If sites serve both Cross-Origin-Opener-Policy: same-origin and Cross-Origin-Embedder-Policy: require-corp, they are now able to use SharedArrayBuffer and Wasm threading again.

It sounds like it should be now possible to enable it via COOP/COEP _instead of_ a flag. Do you have Safari 15.2 to test that assumption on https://webassembly.org/roadmap perhaps?

chicoxyzzy:
It doesn't work by default on https://webassembly.org/roadmap/ and https://alex-wasm.appspot.com/threads/pthread.html

chicoxyzzy:
That's interesting, because SAB works in 15.2

RReverser:
Ugh, so I guess they enabled SAB but not Wasm atomics? Odd choice, seems like such a small step that would've finally completed the Wasm threading journey...

RReverser (approved):

RReverser:
Thanks again :)

RReverser:
Wait, if SAB works, I'm guessing we can at least remove `__XPC_JSC_useSharedArrayBuffer=1` from required flags and only keep the `__XPC_JSC_useWebAssemblyThreading=1`?

RReverser:
@chicoxyzzy can you please check if tests pass with just `__XPC_JSC_useWebAssemblyThreading`?

RReverser:
I've tentatively removed `__XPC_JSC_useSharedArrayBuffer=1 ` from the required flags, hopefully that guess is correct.

chicoxyzzy:
I won't be able to test that this week most likely, will try that on Monday or Tuesday.

PR #260, opened 2022-02-01
Update header and logos

danielnazer:
The header on the homepage is a few years out of date as it has old Firefox and Edge logos and refers to four major browser engines. 

To make the header accurate, these edits change it to "all major browser engines", remove the out-of-date Edge logo, and update the Firefox logo. This also aligns the page with https://webassembly.org/roadmap/, which only shows three engines.


RReverser:
Sure, but now it says a different thing. The point of that statement - at least back then - was specifically to show that all major browser engines implemented 1.0 at the moment of its release. This change instead focuses on where WebAssembly (not necessarily 1.0 / MVP anymore) is implemented in modern engines.

I guess it makes sense to show modern information to new visitors to the website, just saying it loses the original context.

tomayac:
This section was since removed from the homepage, so the PR is now no longer applicable. Closing.

PR #261, opened 2022-02-03
Update to Firefox 96

lars-t-hansen:
We've enabled a bunch of things in FF98 Nightly, and in any case the current FF version is 96.

### DIFF
--- a/features.json
+++ b/features.json
@@ -50,6 +50,11 @@
 			"url": "https://github.com/WebAssembly/reference-types/blob/master/proposals/reference-types/Overview.md",
 			"phase": 4
 		},
+		"relaxedSimd": {
+			"description": "Extensions to fixed-width SIMD with local nondeterminism",
+			"url": "https://github.com/WebAssembly/relaxed-simd/tree/main/proposals/relaxed-simd",
+			"phase": 2
+		},
 		"saturatedFloatToInt": {
 			"description": "Non-trapping float-to-int conversions",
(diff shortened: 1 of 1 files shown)


RReverser:
> and in any case the current FF version is 96

We generally don't update versions to current for any engines, unless they stabilize some feature. It makes it easier for the user to see "ah, so nothing has changed since version X".

Ideally we'd go even a step further and have versions on individual features, but I'm not one to sign-up for maintaining _that_ 😅

Anyway, I'd keep version at 90 for now as it seems like none of the stable features have changed, but I'll happily add the nightly changes.

lars-t-hansen:
> Anyway, I'd keep version at 90 for now as it seems like none of the stable features have changed, but I'll happily add the nightly changes.

Yes, I would prefer to signal that Nightly has these changes.  I'll tweak the PR.

features.json
```diff
@@ -50,6 +50,11 @@
 			"url": "https://github.com/WebAssembly/reference-types/blob/master/proposals/reference-types/Overview.md",
 			"phase": 4
 		},
+		"relaxedSimd": {
+			"description": "Extensions to fixed-width SIMD with local nondeterminism",
```
RReverser, line 54: Hmm, one more thing: `description` is actually the human-readable name that goes into the first column on https://webassembly.org/roadmap/.

I'm generally trying to keep it short where possible, otherwise it will shift all the other columns further out of the viewport.

How about renaming this to just "Relaxed SIMD"?

RReverser (approved):

RReverser:
Thanks!

PR #291, opened 2022-10-05
CSS: prevent vertical scrolling of feature table on some Firefox configurations

keithw:
After the recent refactor (#283), the roadmap table shows a vertical scrollbar in some situations. This PR (allowing the table margin to be inherited) seems to fix the problem and doesn't seem to cause problems in Chrome or elsewhere.

E.g., here's Firefox on Ubuntu/X11 at the default 100% zoom level before this change: image

@andylizi

### DIFF
--- a/css/custom.css
+++ b/css/custom.css
@@ -1959,7 +1959,7 @@ pre code::before, pre code::after {
 }
 
 #feature-support-scrollbox {
-  overflow: auto;
+  overflow: auto hidden;
   margin: 0 0 16px 0;
 }
 


sbc100:
Should we add a column here for wasm2c @keithw ?  Or do you think its maybe too niche?

keithw:
I definitely would like to add a column for wasm2c but was waiting to fix the existing issue (and learn how the website works) before submitting that one...

andylizi:
Hmm, I can't seem to reproduce this on Firefox Windows, which is a bit strange. This PR will give an extra margin between the table and the reference list, which I would prefer to avoid. From what I can tell, there should be only two possibilities that can cause this issue:

1. `#feature-support-scrollbox` somehow got a smaller height than it's inner element `#feature-support`, so the vertical scrollbar appeared. But that doesn't make sense because there ain't any properties like `height` on the `scollbox` so it should automatically fit the content height?
2. `#feature-support` has `overflow: auto`, so maybe the scrollbar came from that element instead of the outside `scrollbox`? I'm also not sure how that could happen. But `overflow: auto` is probably unnecessary here anyways (I must've forgot to remove it).

(Actually, now that I think of it, why do we need an vertical scrollbar here at all :joy:)

So since I can't reproduce locally, can you try the following changes and see if that fixes the issue? 

```diff
 table {
   /* https://bugzilla.mozilla.org/show_bug.cgi?id=1005271 */
   /* display: block; */
   width: 100%;
-  overflow: auto;
 }
```
```diff
 #feature-support-scrollbox {
-  overflow: auto;
+  overflow: auto visible;
   margin: 0 0 16px 0;
 }
```

keithw:
I'm afraid that doesn't seem to fix it -- same as the screenshot above. FWIW this is Firefox 105.0.1 with the "Always show scrollbars" setting enabled. (Without that checked, the vertical scrollbar still appears but only when hovering over the table.)

andylizi:
I'm using Firefox Developer Edition 106.0b7. This is what I see with "Always show scrollbars" on:



Can you screenshot a setup like this (for both `#feature-support-scrollbox` and `#feature-support`)? I'm interested in the "wireframe" and the Box Model tab on the bottom right.

keithw:
Sure, here you go. This is on the current deployed site:

image image

andylizi:
Thanks! Honestly I'm a bit stumped at this point. I booted up Ubuntu Live 22.04.1 with Firefox 105.0.1 with "Always show scrollbars" on, still couldn't reproduce.



Is it possible that this may be caused by an add-on or such?

keithw:
Hmm, if it's easy for you, do you mind trying Xfce/Xubuntu on that VM? This is what I'm running. It may be as simple as "sudo apt install xubuntu-desktop" and then boot the Xubuntu option when logging in. The problem happens even with a fresh install of Firefox, so I don't think it's related to any add-ons.

andylizi:
My steps:
1. Go to https://xubuntu.org/download/ and download `xubuntu-22.04.1-desktop-amd64.iso`.
2. In Hyper-V settings, set the DVD image to the ISO.
3. Boot up the VM, select "Try Xbuntu"
4. In the start menu, click "Web Browser"
5. Go to webassembly.org/roadmap, nothing.
6. Open settings, enable "Always show scrollbars."
7. Restart the browser, still nothing.
8. Firefox version is 103.0.1. Tried `apt update && apt upgrade firefox`, but it says "firefox is already the newest versio (1:1snap1-0ubuntu2)"

keithw:
This is Firefox 105.0.1 (now 105.0.2) direct from Mozilla rather than the Snap. In any event, I just force-pushed a different fix that also seems to work and maybe is more generally palatable. How do you feel about this approach instead?

andylizi:
I don't feel strongly about `hidden` or `visible` either way—it shouldn't matter in our case since the viewport is already vertically scrollable. Though I'm curious about why `visible` wouldn't work, but I guess that's the part of the mystery……

PR #301, opened 2023-01-12
Update Chrome's support of type reflection, v2

gahaas:
Type reflection is available in Chrome and Node.js behind a flag.

PR #303, opened 2023-01-13
Validate `features.json` during CI

andylizi:
To avoid situations like https://github.com/WebAssembly/website/pull/301#discussion_r1068352138 happening again.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: Build & deploy
+name: CI
 
 on:
   push:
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -7,21 +7,49 @@ on:
...
+      run: npm install -g ajv-cli@5 ajv-formats@2
+
+    - name: Validate features table
+      run: ajv -c ajv-formats -s features.schema.json -d features.json --strict=true --errors=text
+
+  build:
+    name: Build & Deploy
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
```
RReverser, line 36: Why is write permission necessary now?

andylizi: https://github.com/peaceiris/actions-gh-pages/commit/414c92c5ff594b380dd8df46dffbb34e3dd2343f

RReverser: If it's the default, then why is this change necessary?

andylizi: It is not necessary. I did this for… symmetry I guess, because I've also set `permissions: {}` for the test job. And I did _that_, because I'm executing arbitrary code from the Internet (`npm install ajv`) so limiting the permission removes the (albeit extremely minuscule) risk of supply chain attacks. There're no drawbacks, and something something the principle of least privilege, so I thought why not.

andylizi:
> To avoid situations like [#301 (comment)](https://github.com/WebAssembly/website/pull/301#discussion_r1068352138) happening again.

[It happened again](https://github.com/WebAssembly/website/pull/313#pullrequestreview-1288329821) :sweat_smile:

PR #309, opened 2023-01-29
Add minimum version for node bigint integration

hoodmane:
Node first supported bigint integration in version 16.0, or in version 14.x and 15.x with `--experimental-wasm-bigint`.

### DIFF
--- a/features.json
+++ b/features.json
@@ -182,7 +182,7 @@
 			"url": "https://nodejs.org/",
 			"logo": "/images/nodejs.svg",
 			"features": {
-				"bigInt": true,
+				"bigInt": "16.0",
 				"bulkMemory": "12.5",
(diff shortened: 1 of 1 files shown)


chicoxyzzy (approved):

chicoxyzzy:
According to https://nodejs.org/en/blog/release/v15.0.0/ it's avaliable since Node 15

chicoxyzzy:
@hoodmane would you be open to update the Node.js version to 15.0?

hoodmane:
Right. I just checked and what confused me is that 15.0 accepts the `--experimental-wasm-bigint` flag but it does nothing, whereas in 16.0 passing this flag is an error. Thanks for catching the mistake.

PR #312, opened 2023-01-31
Move Extended Const Expressions to phase 4

chicoxyzzy:
per today's Wasm CG meeting


chicoxyzzy:
fixed in https://github.com/WebAssembly/website/pull/319

PR #316, opened 2023-02-08
Kotlin: Link to new Wasm target

hfhbd:
Before 1.8.20-Beta (released 2023-02-08), wasm was implemented as a native target using llvm. 1.8.20 contains an experimental new compiler backend.

### DIFF
--- a/getting-started/developers-guide.md
+++ b/getting-started/developers-guide.md
@@ -15,7 +15,7 @@ layout: getting-started
   - Go
     - [with full language support](https://github.com/golang/go/wiki/WebAssembly#getting-started)
     - [targeting minimal size](https://tinygo.org/docs/guides/webassembly/)
-  - [Kotlin](https://kotlinlang.org/docs/reference/native-overview.html)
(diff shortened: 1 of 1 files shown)


getting-started/developers-guide.md
```diff
@@ -15,7 +15,7 @@ layout: getting-started
   - Go
     - [with full language support](https://github.com/golang/go/wiki/WebAssembly#getting-started)
     - [targeting minimal size](https://tinygo.org/docs/guides/webassembly/)
-  - [Kotlin](https://kotlinlang.org/docs/reference/native-overview.html)
+  - [Kotlin](https://kotlinlang.org/docs/whatsnew-eap.html#new-kotlin-wasm-target)
```
chicoxyzzy, line 18: @bashor does it look good to you?

bashor: Thanks! Looks good, but https://kotl.in/wasm would be better, so later, we can change the redirect target :)

hfhbd: Sure, wasn't aware of the redirect.

chicoxyzzy (approved):

PR #390, opened 2024-07-18
Track and update all the implemented features

CryZe:
After seeing https://labs.leaningtech.com/blog/branch-hinting yesterday, I was quite surprised to read that not only is this proposal done, but it's also implemented in all the browsers, and yet it's not even tracked in the `features.json` file.

So I took it upon myself to make sure that all features that are in at least phase 3 (implementation phase) are tracked and up to date. While doing this I however noticed that with the recent addition of "Custom Page Sizes", at least one phase 1 feature is also tracked now. So I exhaustively updated the `features.json` file to include all features that are implemented in at least one engine as well.

With there being a lot more proposals that are being tracked now, I thought it would also make sense to properly reflect the phases on the website, instead of splitting them into just two sections.

### DIFF
--- a/features.js
+++ b/features.js
@@ -1,19 +1,17 @@
 (async () => {
   'use strict';
 
-  function partitionArray(arr, condition) {
-    const matched = [];
-    const unmatched = [];
+  function partitionByPhase(features) {
+    const map = [];
 
-    for (const item of arr) {
-      if (condition(item)) {
-        matched.push(item);
-      } else {
-        unmatched.push(item);
+    for (const feature of features) {
+      if (!map[feature.phase]) {
+        map[feature.phase] = [];
       }
+      map[feature.phase].push(feature);
     }
 
-    return { matched, unmatched };
+    return map;
   }
 
   function h(name, props = {}, children = []) {
@@ -76,14 +74,16 @@
     tBody
   );
 
-  let featureGroups = partitionArray(
+  let featureGroups = partitionByPhase(
     Object.entries(features).map(([name, feature]) => Object.assign(feature, { name })),
-    feature => feature.phase >= 4
   );
 
   featureGroups = [
-    { name: 'Standardized features', features: featureGroups.matched },
-    { name: 'In-progress proposals', features: featureGroups.unmatched },
+    { name: 'Phase 5 - The Feature is Standardized', features: featureGroups[5] },
+    { name: 'Phase 4 - Standardize the Feature', features: featureGroups[4] },
+    { name: 'Phase 3 - Implementation Phase', features: featureGroups[3] },
+    { name: 'Phase 2 - Proposed Spec Text Available', features: featureGroups[2] },
+    { name: 'Phase 1 - Feature Proposal', features: featureGroups[1] },
   ];
 
   // Collect all notes and assign an index to each unique item
@@ -120,6 +120,10 @@
   const columnCount = 2 + Object.keys(browsers).length;
 
   for (const { name: groupName, features } of featureGroups) {
+    if (!features) {
+      continue;
+    }
+
     tBody.append(
       h('tr', {}, [
         h('th', {
--- a/features.js
+++ b/features.js
@@ -1,19 +1,6 @@
 (async () => {
   'use strict';
 
-  function partitionByPhase(features) {
-    const map = [];
-
-    for (const feature of features) {
-      if (!map[feature.phase]) {
-        map[feature.phase] = [];
-      }
-      map[feature.phase].push(feature);
-    }
-
-    return map;
-  }
-
   function h(name, props = {}, children = []) {
     const node = Object.assign(document.createElement(name), props);
     node.append(...children);
@@ -74,8 +61,9 @@
     tBody
   );
 
-  let featureGroups = partitionByPhase(
+  let featureGroups = Object.groupBy(
     Object.entries(features).map(([name, feature]) => Object.assign(feature, { name })),
+    f => f.phase,
   );
 
   featureGroups = [
--- a/features.js
+++ b/features.js
@@ -72,6 +72,7 @@
     { name: 'Phase 3 - Implementation Phase', features: featureGroups[3] },
     { name: 'Phase 2 - Proposed Spec Text Available', features: featureGroups[2] },
     { name: 'Phase 1 - Feature Proposal', features: featureGroups[1] },
+    { name: 'Deprecated', features: featureGroups["deprecated"] },
   ];
 
   // Collect all notes and assign an index to each unique item
--- a/features.json
+++ b/features.json
@@ -16,16 +16,6 @@
 			"url": "https://github.com/WebAssembly/bulk-memory-operations/blob/master/proposals/bulk-memory-operations/Overview.md",
 			"phase": 5
 		},
-		"compilationHints": {
-			"description": "Compilation Hints",
-			"url": "https://github.com/WebAssembly/compilation-hints/blob/main/proposals/compilation-hints/Overview.md",
-			"phase": 1
-		},
-		"componentModel": {
-			"description": "Component Model",
-			"url": "https://github.com/WebAssembly/component-model",
-			"phase": 1
-		},
 		"customAnnotationSyntaxInTheTextFormat": {
 			"description": "Custom Text Format Annotations",
 			"url": "https://github.com/WebAssembly/annotations/blob/main/proposals/annotations/Overview.md",
@@ -34,7 +24,7 @@
 		"customPageSizes": {
 			"description": "Custom Page Sizes",
 			"url": "https://github.com/WebAssembly/custom-page-sizes/blob/main/proposals/custom-page-sizes/Overview.md",
-			"phase": 1
+			"phase": 2
 		},
 		"esmIntegration": {
 			"description": "ESM integration",
@@ -71,20 +61,10 @@
 			"url": "https://github.com/WebAssembly/js-string-builtins/blob/main/proposals/js-string-builtins/Overview.md",
 			"phase": 3
 		},
-		"halfPrecision": {
-			"description": "Half Precision",
-			"url": "https://github.com/WebAssembly/half-precision/blob/main/proposals/half-precision/Overview.md",
-			"phase": 1
-		},
 		"memory64": {
 			"description": "Memory64",
 			"url": "https://github.com/WebAssembly/memory64/blob/master/proposals/memory64/Overview.md",
 			"phase": 3
-		},
-		"memoryControl": {
-			"description": "Memory control",
-			"url": "https://github.com/WebAssembly/memory-control/blob/main/proposals/memory-control/Overview.md",
-			"phase": 1
 		},
 		"multiMemory": {
 			"description": "Multiple memories",
@@ -101,11 +81,6 @@
 			"url": "https://github.com/WebAssembly/mutable-global/blob/master/proposals/mutable-global/Overview.md",
 			"phase": 5
 		},
-		"referenceTypedStrings": {
-			"description": "Reference-Typed Strings",
-			"url": "https://github.com/WebAssembly/stringref/blob/main/proposals/stringref/Overview.md",
-			"phase": 1
-		},
 		"referenceTypes": {
 			"description": "Reference types",
 			"url": "https://github.com/WebAssembly/reference-types/blob/master/proposals/reference-types/Overview.md",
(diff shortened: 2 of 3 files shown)


features.json
```diff
@@ -2,60 +2,105 @@
...
-			"description": "Custom page sizes",
+			"description": "Custom Page Sizes",
 			"url": "https://github.com/WebAssembly/custom-page-sizes/blob/main/proposals/custom-page-sizes/Overview.md",
 			"phase": 1
 		},
+		"esmIntegration": {
+			"description": "ESM integration",
+			"url": "https://github.com/WebAssembly/esm-integration",
+			"phase": 3
+		},
```
CryZe, line 43: This is the only one that's not implemented by any, but is in phase 3, which falls into my "everything phase 3 and up having to be listed" rule.

features.json
```diff
@@ -103,36 +168,50 @@
...
 			"features": {
 				"bigInt": "78",
+				"branchHinting": ["flag", "Enabled in Nightly, unavailable in Beta/Release"],
 				"bulkMemory": "79",
+				"customAnnotationSyntaxInTheTextFormat": null,
 				"exceptions": "100",
 				"extendedConst": "112",
 				"gc": "120",
 				"jspi": ["flag", "Requires flag `javascript.options.wasm_js_promise_integration` in Nightly, unavailable in Beta/Release"],
+				"jsStringBuiltins": ["flag", "Enabled in Nightly, requires flag in Beta/Release"],
```
CryZe, line 212: This one is weird, it's somehow also available on Beta / Release:  https://bugzilla.mozilla.org/show_bug.cgi?id=1883921

CryZe: I updated it, and some others with the name of the flags.

features.js
```diff
@@ -1,19 +1,17 @@
...
+    for (const feature of features) {
+      if (!map[feature.phase]) {
+        map[feature.phase] = [];
       }
+      map[feature.phase].push(feature);
     }
 
-    return { matched, unmatched };
+    return map;
   }
```
CryZe, line 15: The JS changes I made have not been tested btw, not sure how well the CI works here.

features.json
CryZe: Nvm, turns out it's implemented by Webkit.

features.json
```diff
@@ -2,60 +2,105 @@
 	"$schema": "./features.schema.json",
 	"features": {
 		"bigInt": {
-			"description": "JS BigInt to Wasm i64 integration",
+			"description": "JavaScript BigInt to WebAssembly i64 integration",
```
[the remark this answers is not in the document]
CryZe: I reverted some again, because it made the table not fit anymore.

features.js
CryZe: I built and verified it locally, so this is no longer a concern.

CryZe:
Here is what the new table looks like:
localhost_8000_features_ (1)

features.json
dschuff: Interesting, I wasn't aware of this. Do you know if it's up-to-date? There have been quite a few changes in the long time since the proposal started. Maybe @guybedford knows.
If it's the original version of the proposal and not the current one, we probably shouldn't include it.

features.json
```diff
@@ -6,56 +6,106 @@
...
-			"phase": 4
+			"phase": 5
 		},
 		"mutableGlobals": {
-			"description": "Mutable globals",
+			"description": "Import/Export of Mutable Globals",
 			"url": "https://github.com/WebAssembly/mutable-global/blob/master/proposals/mutable-global/Overview.md",
 			"phase": 5
 		},
+		"referenceTypedStrings": {
```
dschuff, line 104: stringref is implemented in V8, but it's not under active development anymore, and might disappear. I'm not sure I would want to include it here.

features.json
```diff
@@ -64,15 +114,20 @@
...
 			"url": "https://github.com/WebAssembly/relaxed-simd/tree/main/proposals/relaxed-simd",
-			"phase": 4
+			"phase": 5
 		},
 		"saturatedFloatToInt": {
 			"description": "Non-trapping float-to-int conversions",
 			"url": "https://github.com/WebAssembly/spec/blob/master/proposals/nontrapping-float-to-int-conversion/Overview.md",
 			"phase": 5
 		},
+		"sharedEverythingThreads": {
```
dschuff, line 124: This is another one where the spec is in such an early state, and the implementation in V8 is very early and just reflects a small part of what's in the spec.  I don't think I would want to claim in this table (which has no room for nuance) that the spec is "implemented"

dschuff (commented):
This is impressive!
One concern I have is that this table is meant to help developers who might not be following super-closely along with the progress of specs, and might want to know whether they should try out a feature (or target it with a compiler). For very early-stage proposals, the answer is usually "maybe; it's complicated". I'm not sure it's helpful to just declare that a feature is supported, even with a flag. I don't know if that means we should just make a judgement call about whether to include a feature, or have some other caveats, or treat stage 1 proposals differently, or what.

rossberg:
This is very nice. But I have to agree with @dschuff about early phases: there is basically no entry requirement to phase 1, and proposals at this stage generally are "just an idea", often with no worked out design that could even be implemented and no agreement yet that they should ever be included. Implementations at this stage are mostly experiments. So it may be misleading to show phase 1 in this table, and may set wrong expectations. In contrast, phase 2 at least requires a "complete overview document" and "reasonably high level of consensus".

features.js
```diff
@@ -1,19 +1,17 @@
...
-  function partitionArray(arr, condition) {
-    const matched = [];
-    const unmatched = [];
-
-    for (const item of arr) {
-      if (condition(item)) {
-        matched.push(item);
-      } else {
-        unmatched.push(item);
+  function partitionByPhase(features) {
```
andylizi, line 4: I can't comment on the status of proposals, but on the technical side of things, perhaps we can use [`Object.groupBy`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy) here. It's a rather new addition, but I think the assumption was that [developers would use up-to-date browsers](https://github.com/WebAssembly/website/pull/283#issuecomment-1234044185) anyway.

features.json
```diff
@@ -6,56 +6,106 @@
...
+			"phase": 1
+		},
+		"customAnnotationSyntaxInTheTextFormat": {
+			"description": "Custom Text Format Annotations",
+			"url": "https://github.com/WebAssembly/annotations/blob/main/proposals/annotations/Overview.md",
+			"phase": 5
+		},
 		"customPageSizes": {
-			"description": "Custom page sizes",
+			"description": "Custom Page Sizes",
```
andylizi, line 35: Most titles on the table currently use Title Case, but there are still a few sentence cases mixed in, like "ESM integration" below. This may be a good opportunity to unify them.

rossberg:
Custom page sizes is scheduled for a phase 2 vote next Tuesday, so we could just wait that out. The component model is rather special, since it is it's own separate layer and will not necessarily be implemented in engines.

features.json
dschuff: Most (but not all) of the proposals listed before this PR only had the first word capitalized. https://github.com/WebAssembly/proposals mostly (but not fully) uses title case. I don't have a strong opinion about which, but maybe just switch the missing few here to title case for consistency would be good.

dschuff:
That's a good point about the phase 2 requirements being a good proxy for the usefulness of trying out a feature. I think that makes sense as a "default" criterion for inclusion here.
I'm fine leaving custom page sizes in for now though, even if we don't wait; if it doesn't make it to phase 2, we can figure out what to do.
CM is more interesting. My intuition is that it's well-specified enough that it might be useful for users or tool-builders to try to use, but as @rossberg mentioned, the layering is complicated and the relevant implementations are mostly not in this table. I'd say it probably makes sense to leave it out of this table for this PR, but if there's interest in coming up with an analogous convenience for developers, we can certainly discuss that.
/cc @lukewagner

features.json
CryZe: I updated it to use title case everywhere.

CryZe:
I removed all the phase 1 proposals and preemptively moved `Custom Page Sizes` to phase 2.

features.js
CryZe: I'll look into that tomorrow, thanks.

features.json
CryZe: I'd guess it's not up to date, but I don't actually know.

features.js
CryZe: Alright, this is now done.

CryZe:
Here is the latest table, with most concerns addressed: 
localhost_8000_features_ (3)

features.json
```diff
@@ -2,77 +2,102 @@
...
+			"phase": 2
+		},
+		"esmIntegration": {
+			"description": "ESM Integration",
+			"url": "https://github.com/WebAssembly/esm-integration",
+			"phase": 3
 		},
 		"exceptions": {
-			"description": "Exception handling",
+			"description": "Exception Handling",
```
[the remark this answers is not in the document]
dschuff: Ah, i just noticed that you describe that in the notes in this PR. My concern with this approach is that developers might think there is no supported EH option across the web, while in fact there is (i.e. until exnref is supported on all browsers, I expect that at least developers of C++ and any other toolchain that supports the existing EH will continue to use that).
Maybe we should figure out what flag options we want to use in emscripten and revisit this...
But anyway.... to avoid letting the perfect being the enemy of the good and stalling this too long, maybe let's just land this PR and then split it out in a separate PR. 
So maybe my question for you here is, how easy would it be with this code architecture to add another category to the table for deprecated features?

dschuff (commented):
Sorry for more comments! Now you're discovering why some of these were "conveniently" left out :D
thanks for perservering!

features.json
```diff
@@ -103,65 +138,82 @@
...
+				"typeReflection": ["flag", "Requires flag `--enable-wasm-type-reflections`"],
+				"webContentSecurityPolicy": "102"
 			}
 		},
 		"Safari": {
 			"url": "https://www.apple.com/safari/",
 			"logo": "/images/safari.svg",
 			"features": {
 				"bigInt": ["15", "wasm-bigint is supported in desktop Safari since 14.1 and iOS Safari since 14.5; however BigInt64Array, which is needed by Emscripten, was released in 15"],
+				"branchHinting": "16",
```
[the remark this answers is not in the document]
CryZe: It has been added in Safari 16, and was immediately enabled by default. There is now no longer any feature to toggle it. I'll recheck to see if they "removed the feature" or if that's just because they don't need a toggle anymore.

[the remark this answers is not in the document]
CryZe: Yep, they removed the flag because it's "already shipped" here: https://github.com/WebKit/WebKit/commit/ca9f4e631e5eaf44ce22e1f5cc29519be9488825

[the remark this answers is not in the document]
dschuff: aha, and it was enabled in https://github.com/WebKit/WebKit/commit/e1f19d73022eeb9d97b02fd64ba45e40d346d169
I'll bet this was an accident, actually. TC39 has a similar stage process for JS features, and IIRC browsers typically ship at stage 3. Anyway, this is fine for the website, I'll follow up and make sure there's nothing that they need to fix or update with the proposal :)

features.json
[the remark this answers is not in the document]
CryZe: I split it into a deprecated section now.

features.json
```diff
@@ -82,19 +107,29 @@
...
+			"url": "https://github.com/WebAssembly/function-references/blob/main/proposals/function-references/Overview.md",
+			"phase": 5
+		},
 		"typeReflection": {
-			"description": "Type reflection",
+			"description": "Type Reflection",
 			"url": "https://github.com/WebAssembly/js-types/blob/main/proposals/js-types/Overview.md",
 			"phase": 3
+		},
+		"webContentSecurityPolicy": {
```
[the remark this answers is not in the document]
CryZe: I removed CSP for Node and Deno.

features.json
```diff
@@ -2,77 +2,107 @@
...
+			"phase": 3
+		},
+		"exceptionsFinal": {
+			"description": "Exception Handling",
+			"url": "https://github.com/WebAssembly/exception-handling/blob/master/proposals/exception-handling/Exceptions.md",
+			"phase": 4
 		},
 		"exceptions": {
-			"description": "Exception handling",
+			"description": "Exception Handling",
```
dschuff, line 40: How about we call this "Legacy Exception Handling" to match the spec, and for now how about calling `exceptionsFinal` "Exception Handling with exnref` to disambiguate and for consistency with Emscripten. Once some time has passed and the emscripten/LLVM implementation is mature we can start changing the language in the documentation to make exnref more like the "default".

dschuff (commented):
with this change the PR LGTM

features.json
CryZe: Alright, I renamed them.

CryZe:
I pushed one additional change marking GC as being enabled in Safari 18. While it is indeed now enabled by default in Webkit, I don't actually know whether that also means it will be in the final Safari 18 or they ship an earlier version of Webkit.

CryZe:
I undid that change. It seems like they mostly just cherry pick changes during the beta period, so it might not make it until 18.1. I'll just do a follow up PR then.

dschuff:
> I undid that change. It seems like they mostly just cherry pick changes during the beta period, so it might not make it until 18.1. I'll just do a follow up PR then.

I think this is the right thing to do. When a feature turns on by default in WebKit it *usually* means it will ship in Safari, but not always. And Apple is very opaque about their release process and will never say anything about what will release in the future. So in my experience it's just best to wait until the official beta releases come out.

dschuff:
This is great, thanks for all your work on this!

PR #411, opened 2024-12-11
Fix flag for js string builtins

peter-jerry-ye:
Run 
```bash
deno --v8-flags=--help
node --help --v8-options
```

And there's no `experimental-wasm-js-string-builtins`. Instead, it's ` --experimental-wasm-imported-strings`


peter-jerry-ye:
Also notice that it's default on for deno since, I don't know, at least 2.1.3?

tomayac (approved):
LGTM, thank you

Issue #429, opened 2025-02-18
Make feature extensions support table more readable


remexre:
Right now, it's constrained to the same column width as the prose; instead, it should expand to the width available.

The top and left headers should both be "sticky" when it's being scrolled vertically or horizontally as well.

PR #430, opened 2025-02-19
Allow feature table to expand outside content area

andylizi:
Partially address #429.

Preview

### DIFF
--- a/css/custom.css
+++ b/css/custom.css
@@ -2001,11 +2001,6 @@ header {
   text-align: center;
 }
 
-.container img,
-.container-narrow img {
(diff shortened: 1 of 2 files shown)


keithw:
The wasm2c logo seems to be getting squeezed -- it should be scaled proportionately like the others.

andylizi:
> The wasm2c logo seems to be getting squeezed -- it should be scaled proportionately like the others.

Thanks, should be fixed now!

tomayac (approved):
LGTM, thank you

I'll address the sticky headers part in a separate commit.

PR #432, opened 2025-02-19
Compilation Hints are now in Phase 2

CryZe:
This additionally updates Wide Arithmetic to Phase 3. Also Deno now supports JS String Builtins.


tomayac (approved):
LGTM, thank you!

PR #436, opened 2025-03-15
Update feature flag for stack switching proposal

pufferfish101007:
It seems that the v8 flag listed on https://webassembly.org/roadmap for the stack switching proposal is incorrect; the `--experimental-wasm-stack-switching` flag exists, but it [seems to be unused](https://github.com/search?q=repo%3Av8%2Fv8+path%3Asrc%2Fwasm+stack_switching&type=code) (note that all references to stack_switching flags are different v8 flags relating to jspi [I think]). However, the `--experimental-wasm-wasmfx` flag [has been added recently](https://chromium-review.googlesource.com/c/v8/v8/+/6158267), with an incomplete implementation (not yet landed in stable chrome as far as I can tell).

#400 says that there was an in-progress implementation of stack switching back in September, but as far as I can tell this isn't correct?

This pull request updates the listed feature flags for the stack switching proposal... it's difficult to test these because the change was made recently enough to not be in stable releases yet, but hopefully it's correct now. (And hopefully I haven't completely misunderstood the whole situation.)

### DIFF
--- a/features.json
+++ b/features.json
@@ -195,7 +195,7 @@
         "simd": "91",
         "stackSwitching": [
           "flag",
-          "Requires CLI flag `--js-flags=--experimental-wasm-stack-switching`"
+          "Requires CLI flag `--js-flags=--experimental-wasm-wasmfx`"
         ],
         "tailCall": "112",
         "threads": "74",
@@ -335,7 +335,7 @@
         "simd": "16.4",
         "stackSwitching": [
           "flag",
-          "Requires flag `--experimental-wasm-stack-switching`"
+          "Requires flag `--experimental-wasm-wasmfx`"
         ],
         "tailCall": "20.0",
         "threads": "16.4",
@@ -389,7 +389,7 @@
         "simd": "1.9",
         "stackSwitching": [
           "flag",
-          "Requires flag `--v8-flags=--experimental-wasm-stack-switching`"
+          "Requires flag `--v8-flags=--experimental-wasm-wasmfx`"
         ],
         "tailCall": "1.32",
         "threads": "1.9",


tomayac:
Thanks for the PR, @pufferfish101007! @tlively Do you know who from V8 would be a good person to verify this? Likely @fgmccabe!?

fgmccabe:
The experimental-wasm-stack-switching was the one originally used to enable JSPI. We implemented the experimental-wasm-jspi flag to avoid confusion; and we kept the original flag as a synonym.
The experimental-wasm-wasmfx flag is there to guard the prototype implementation of core stack switching. Very much a 'work in progress' at the moment.

Liedtke:
@fgmccabe Note that the [flag is documented](https://source.chromium.org/chromium/chromium/src/+/main:v8/src/wasm/wasm-feature-flags.h;l=45;drc=f111bd3e66ed81a0115562282df0c47ef68567a1) to be used for the stack switching proposal though:
```
  /* Stack Switching proposal. */                                              \
  /* https://github.com/WebAssembly/stack-switching */                         \
  /* V8 side owner: thibaudm, fgm */                                           \
  V(stack_switching, "stack switching", false)                                 \
  ```
Should we update the comment then to avoid further confusion? (I was confused by this as well a few weeks ago.)

fgmccabe:
We could. Not sure it is worth it though:
We are pretty close now to shipping JSPI. 
Originally, it seemed likely that we would have our own design proposal, and having a separate flag for that seemed useful. Now, that scenario is unlikely: there is a unified proposal that is based on the wasm-fx proposal.
At some point after shipping JSPI we will be removing the JSPI flag; we can remove wasm-stack-switching at the same time.

tomayac:
What's the verdict here? Should we close this PR for now?

tomayac:
@Liedtke Could you make the change, as you're close? Happy to merge once you tell me. Thanks!

Liedtke:
@tomayac Sure, I assume that means the right thing is to remove the flag mention for all V8-based implementations for now: https://github.com/WebAssembly/website/pull/444

tomayac:
#444 is merged now, so what do we do with this present PR? 

> If "incomplete" features behind a flag should display their flag, I guess, this PR can just be merged

So, can it? (Sorry, I'm not involved enough to have an answer.)

tomayac:
We also have a conflict now of course…

Liedtke:
> So, can it? (Sorry, I'm not involved enough to have an answer.)

Well, maybe that was the misunderstanding, based on your request for a PR I interpreted it as "it doesn't make sense to display 'this feature exists behind the following flag' if you can't do anything useful when setting that flag."

Probably we want to "publish" feature flags here once they are in a state where you can start experimenting with them? (Which wouldn't require having the full proposal implemented but at least some part that is usable and testable.)

tomayac:
Alright, so closing this PR then. Thank you!

PR #437, opened 2025-03-18
Add `/news/` section

tomayac:
This adds an **Updates** section to the website. 

- Each update post lives in `/updates/yyyy-mm-dd-slug/` in an `index.md` file and all local resources like images in the same folder.

   Screenshot 2025-03-18 at 19 27 36
- I didn't manage to make the slug work in the `/updates/` landing page. I think somehow Jekyll doesn't actually recognize my collection. As a current workaround, the raw path works. Someone more familiar with Jekyll internals may know how to fix this.

   Screenshot 2025-03-18 at 19 28 11

@lukewagner and @rossberg, does this approach work for your (time-critical) needs?


rossberg (commented):
Excellent, thanks a lot!

Just one suggestion: Can we call it "News" instead? That's what I'd be looking for as a visitor used to similar web sites (e.g., bytecodealliance.org, python.org, ocaml.org, ruby-lang.org, etc.). "Updates" otoh can mean different (technical) things and is much less obvious.

_includes/header.html
```diff
@@ -38,6 +38,9 @@
           <a class="site-nav-item btn" href="/community/resources/"
             >Community</a
           >
+          <a class="site-nav-item btn" href="/updates/"
+            >Updates</a
```
tomayac, line 42:
```suggestion
          <a class="site-nav-item btn" href="/news/"
            >News</a
```

tomayac:
Opened https://github.com/WebAssembly/website/issues/438 as a follow-up.

PR #444, opened 2025-03-27
Remove wasm stack switching flag mention for V8

Liedtke:
`--experimental-wasm-stack-switching` is a deprecated flag for the JSPI proposal and not related to stack switching. Stack switching will be implemented behind `--experimental-wasmfx` but is still in a very early stage.

### DIFF
--- a/features.json
+++ b/features.json
@@ -193,10 +193,6 @@
         "saturatedFloatToInt": "75",
         "signExtensions": "74",
         "simd": "91",
-        "stackSwitching": [
-          "flag",
-          "Requires CLI flag `--js-flags=--experimental-wasm-stack-switching`"
-        ],
         "tailCall": "112",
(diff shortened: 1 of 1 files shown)


tomayac (approved):
LGTM

Liedtke:
@tomayac I don't have any merging rights. Note that I also didn't build it locally, how relevant is that? :smile: 
Ruby / bundle is throwing all kinds of permission errors as it wants to write to root owned directories... :face_with_peeking_eye:

tomayac:
No worries, I can merge. Yes, getting the site up and running locally has been a challenge for me as well. And then once I had it running I thought I'd update the dependencies, and then the GitHub deploy broke. Since all we need is the static output, I think it's fine to just live with outdated dependencies for a little longer…

PR #454, opened 2025-05-07
Rename "Feature Extensions" to "Feature Status"

sunfishcode:
\#361 renamed this page from "Future features" to "Feature Extensions', as many of these features are standardized and widely implemented today. Looking at it now, the word "extensions" may also be misleading, as it may be read as suggesting that the MVP remains the base language, with all other features being held separate, as discrete extras.

In practice, when features are added to the spec, they are integrated and not kept separate. They appear as indistinguishable from the features that were in the MVP.

Consequently, I propose this page be titled "Feature Status", as it documents the implementation status of various features.

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -35,7 +35,7 @@
           >
           <a class="site-nav-item btn" href="/specs">Specs</a>
           <a class="site-nav-item btn" href="/docs/faq/">Docs</a>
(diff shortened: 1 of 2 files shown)


tlively (approved):
This makes sense to me. PTAL @tomayac.

tomayac (approved):
LGTM from me, too.

PR #462, opened 2025-07-18
Node.js supports memory64 starting with v24.0

Liedtke:
See https://nodejs.org/en/blog/release/v24.0.0#v8-136


tomayac (approved):
LGTM, thank you!

PR #488, opened 2025-11-16
features: Update list of Phase 1 & 2 proposals

pdubroy:
- Add "JIT Interface" and "Multibyte Array Access" to Phase 1
- Update some other proposals to use title case, consistent with other entries


tomayac (approved):
LGTM, thanks for being on top of those!

PR #492, opened 2025-11-26
[feature]: add Owi

redianthus:
Hi,

This is adding [Owi](https://github.com/ocamlpro/owi) to the list of engines.

Best,


tomayac (approved):
LGTM, thank you

Issue #495, opened 2025-12-05
[Features] Binaryen toolchain request


yhx-12243:
In https://webassembly.org/features/ page, several browsers and toolchains that supports wasm are demonstrate whether features are enabled.

However, [`binaryen`](https://github.com/WebAssembly/binaryen) is also a well-known toolchain (even appears many time in this repo!) but not appear in this list.

So this issue is a request to adding https://github.com/WebAssembly/binaryen into [features.json](https://github.com/WebAssembly/website/blob/main/features.json).

dschuff:
Yeah, it was originally just browsers, and then other implementations, and toolchains is a new thing. So it's grown beyond the scope of what it was intended for. Probably we should have had this discussion before we added owi.
I don't necessarily mind having a list of toolchains but if we do that, I think the page needs a redesign. It already has more than a reasonable number of columns and there are a lot of toolchains in the world that do or might eventually support wasm.

kripken:
Yes, perhaps we could redesign the page to have a section for VMs and another for toolchains, or something like that.

tomayac:
For now, since we have consensus to add it, can someone familiar with the details open a PR to add the data to `features.json` and we can then think of a potential redesign later?

PR #497, opened 2025-12-09
Add Binaryen

kripken:
Context: #495 

Per discussion there, we may want to land this after a redesign (perhaps with
separate areas for VMs and toolchains, of which this is the second). I don't have
a feeling of urgency here, either way seems fine, however @dschuff / @tomayac 
decide.


tomayac (approved):
LGTM, thank you

PR #498, opened 2025-12-10
Feature table redesign

andylizi:
### Preview

~~https://wasm-website-preview.pages.dev/features/~~

Merged, now available at https://webassembly.org/features/

### Motivation

1. As discussed in #495, the feature table is continuously growing (more runtimes & tools). Some kind of redesign was needed to keep the column count reasonable.
1. The old implementation (mostly written by me) is plain JS that builds the DOM in an ad-hoc way. It has a lot of moving parts tangled together (feature detection, async loading, tooltip logic w/ timers, footnote mappings, etc). It's becoming frustrating to extend, and it's designed to "generate once" which fits poorly for dynamic behavior. A small templating solution would greatly improve maintainability.
1. The current UI relies on Wikipedia-style tooltips for notes, which I now think are a poor fit:.
	- Every cell has a tooltip that appears on hover and tends to "capture" the cursor when moving between cells.
	- Having a giant list of every note after the table is unwieldy and unnecessary. It reads like a bibliography, and it's growing in length.
	- Semantically speaking, the cells and their notes should be strongly associated, contrary to a "reference" relationship.
	- Keyboard navigation: only cells with footnotes are tabbable, which feels a bit inconsistent. Not sure if this is actually a problem for Assistive Technology users though.

### Design decisions

- **Filtering** (addressing motivation 1): this PR uses checkboxes to toggle column categories (`Web Browsers`, `Standalone Runtimes`, `Tools`). While this does the job for now, I'm not super in love with this design and would welcome ideas.
  - Categories are hard to balance: `Standalone Runtimes` currently dominate while the others are sparse. This would worsen as more runtimes are added.
  - It's tricky to make the categories balanced. Right now, `Standalone Runtimes` have too many columns and the other two too few. We would still run into the columns problem in the future, when there're more server runtimes added.
  - Multiple stacked tables isn't great, the page is already very long vertically.
  - Multiple pages would hurt discoverability and prevent cross-category comparison. And it doesn't address the unbalanced problem.
- **Visual style** (addressing motivation 3): heavily "inspired" by MDN's browser compatibility table ([example](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input#browser_compatibility)). Their design works really well here. I emulated the look & feel without looking at their code, so I think it should avoid license issues? Otherwise, since their site uses MPL, a file-based copyleft license, we can make the relevant files MPL to be extra safe.
- **JS framework** (addressing motivation 2): I used [Alpine.js](https://alpinejs.dev/) as a minimal framework, because it is the only option AFAIK that doesn't require a (JS-based) build step. Our current build process is Ruby-based and hasn't been touched in 8+ years, so I didn't dare to make changes. Let me know if we'd like to explore alternative approaches.

### Current status

~~I'd appreciate early feedback on whether the overall direction looks reasonable. I know large changes can be difficult to review with the project's limited bandwidth. If an incremental path would be preferable, I'm happy to take that route instead. I just don't have a clear idea to do this cleanly under the constraint of the existing code, so suggestions are very welcome.~~ Thanks everyone for the input!

This is ready for review.

Since `feature.js` is a wholesale rewrite, it's recommended to look at the new file rather than the messy diff.

### DIFF
--- a/features.js
+++ b/features.js
@@ -44,24 +44,6 @@ function mapValues(obj, mapper) {
   );
 }
 
-/**
- * Break a string into three parts using the given delimiter.
- * @param {string} str
- * @param {string} delim
- * @returns {[string, string, string]}
- */
-function splitParts(str, delim) {
-  const start = str.indexOf(delim);
-  const end = str.indexOf(delim, start + 1);
-  if (start >= 0 && end > start) {
-    const head = str.substring(0, start);
-    const body = str.substring(start + 1, end);
-    const tail = str.substring(end + 1);
-    return [head, body, tail];
-  }
-  return [str, '', ''];
-}
-
 function loadFeatureDetection() {
   // Please cache bust by bumping the `v` parameter whenever `feature.json` is
   // updated to depend on a new version of the library. See #353 for discussion.
@@ -504,20 +486,27 @@ const state = () => ({
   /** @param {string | undefined} note  */
   renderNote(note) {
     if (!note) return;
+    const fragment = document.createDocumentFragment();
 
     // Transform markdown-like backticks into html <code></code>
-    // TODO: use regex
-    const fragment = document.createDocumentFragment();
-    while (note) {
-      const [head, body, tail] = splitParts(note, '`');
+    const re = /`([^`]+)`/g;
+    /** @type {RegExpExecArray | null} */
+    let match = null,
+      lastIndex = 0;
+    while ((({ lastIndex } = re), (match = re.exec(note)))) {
+      const head = note.substring(lastIndex, match.index),
+        body = match[1];
+
       head && fragment.append(head);
       if (body) {
         const el = document.createElement('code');
         el.textContent = body;
         fragment.appendChild(el);
       }
-      note = tail;
     }
+
+    const tail = note.substring(lastIndex);
+    tail && fragment.append(tail);
     return fragment;
   },
 


redianthus:
It looks very nice!

I think maybe we should have the ability for a project to belong to more than a single category? For instance, Owi features both a "standalone runtime", and some "tooling" (symbolic execution, formatter etc.).

jryans:
There’s still a small issue with the revealed details, at least in mobile iOS browsers:

image

I assume the flag explanation should move to a second line instead of being partially offscreen.

Other that that, looking good to me. 😄

andylizi:
Latest preview: [still the same link](https://wasm-website-preview.pages.dev/features/).

> but there is also a regular, concrete interpreter `owi run`

Ah yes I missed that, thanks for pointing it out! Yeah that makes sense, if nobody objects I'll added Owi to the Standalone Runtimes category as well.

The table now supports specifying multiple categories, with the first one being primary. If both Runtimes and Tools are selected, Owi will be listed under Tools. If Tools is not selected, Owi will be listed under Runtimes. That should hopefully be a consistent way to handle everything.

> As you're touching the CSS anyways, the wazero logo almost disappears in dark mode. Not sure if anyone from wazero reads this and can advise on branding guidelines, but a simple invert for this particular logo would likely do the job.

Done!

> There’s still a small issue with the revealed details, at least in mobile iOS browsers:

Looks like a Safari quirk with SVG sizes, should be fixed now!

---

With the design issues mostly settled, I will start working on polishing and getting the code ready for review. Thanks again for all the feedback!

andylizi:
The code should be ready for review!

There are no visible UI changes since the last commit, but here's the preview: https://wasm-website-preview.pages.dev/features/

tomayac (approved):
LGTM, thank you

This looks super nice and is very useful now. Also love the fact that it's deep-linkable. There will likely be bugs that we encounter once this is live, but for this to happen, it needs to be live, so let's merge now and deal with anything else that might come up in follow up PRs. Thanks again for this heroic effort!

Liedtke:
Indeed, works now (I also checked on mobile to rule out some VPN issue and it was there as well, so I assumed it was some other issue e.g. with relative paths, cross-site-requests or other things that suddenly break when being deployed to the actual URL).

Thanks for checking!

PR #552, opened 2026-07-24
Migrate Chicory to Endive

andreaTP:
Chicory has been rebranded to Endive and moved under the Bytecode Alliance( [ref](https://bytecodealliance.org/articles/endive-and-the-next-chapter-of-webassembly-on-the-jvm) ).

Update the runtime entry with the new name, URL (endive.run), logo, and version numbers. Remove the tail call interpreter-only caveat (now fully supported) and add `extendedConst` support.


tomayac (approved):
LGTM, thank you!

🥗 I see what you did there…

PR #554, opened 2026-08-04
Add GraalVM Web Image.

fniephaus:
GraalVM Web Image is still an experimental feature but maybe it makes sense to mention it to the getting started page already.


tomayac (approved):
LGTM, thanks

PR #562, opened 2026-08-21
Move Wasmi next to Wasm3 + update entries

Robbepop:
Reason: Wasmi and Wasm3 are friendly competitors and thus it is useful to readers of the website to have them be adjacent to each other for comparison.

## The Problem

Back when I added Wasmi to the list I politely appended it to the back.
However, I noticed that not all new additions do that.

The issue is that the last spots are inferior to the first due to the visualization of the table on the website.
Users have to scroll to see the last spots and therefore nobody wants to take those places.

This visualization worked until a certain threshold of entries existed.
How could we make this visualization fairer for now and future additions?

## Some Proposals

- Maybe adding more cathegories such as Web Runtime, JIT runtime, Interpreter runtime, Compilers, etc..?
    - Using tabs (that automatically cycle through) to switch between views so that nobody has to scroll for anything?
- Maybe a completely new visualization that no longer is a table?
- Maybe randomize the order each visit to the website?

### DIFF
--- a/features.json
+++ b/features.json
@@ -667,6 +667,39 @@
         "webContentSecurityPolicy": null
       }
     },
+    "Wasmi": {
+      "url": "https://github.com/wasmi-labs/wasmi/",
+      "logo": "/images/wasmi.png",
+      "category": "Standalone Runtimes",
+      "features": {
+        "bigInt": null,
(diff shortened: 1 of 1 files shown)


tomayac (approved):
LGTM, thanks

PR #563, opened 2026-08-22
Add zwasm to the standalone runtimes table

jtakakura:
Adds [zwasm](https://github.com/zwasm/zwasm), a WebAssembly runtime written in Zig, to the Standalone Runtimes rows, plus its logo (`images/zwasm.svg`).

Every cell was measured against zwasm v2.5.0 (default build) rather than transcribed from project docs: the official spec-testsuite lanes run on both engines (interpreter and JIT), plus a minimal self-verifying probe module per feature. Where the engines differ, the cell carries a note naming which engine executes the feature and what the default mode does.

One question for the maintainers — non-blocking: zwasm ships two engines with an `auto` default that routes between them, so a few cells are capability-split rather than flag-gated: `simd` / `relaxedSimd` (the JIT executes them; the interpreter accepts the module but traps when a vector instruction executes), `multiMemory` (interpreter only; the JIT rejects), and `threads` (single-threaded host: wait times out, notify returns 0). Existing annotations in the table mostly describe how to *enable* a feature, though Wasmtime's `tailCall` note is a backend-and-architecture condition. Are these cells represented the way you would want, or would you rather see them as `false`, or worded some other way? If they are fine as written, nothing needs to change.

Versions are omitted deliberately (plain `true`) — per-feature "supported since" versions are not cheaply derivable from the project history.

Logo: silhouette derived from the WebAssembly logo mark (CC0-1.0, carlosbaraza/web-assembly-logo), recoloured, with lettering set in Open Sans (OFL-1.1).


tomayac (approved):
LGTM, thanks!

PR #566, opened 2026-08-27
Update compact import section

dschuff:
It advanced to phase 4 in August 2026
It's also supported with a flag in Chrome.


tomayac (approved):
LGTM, thanks

PR #571, opened 2026-09-22
Update features.json in a GitHub Actions workflow

gsnedders:
This adds a GitHub Actions workflow, similar to what https://github.com/tc39/dataset does, to keep features.json up-to-date.


package.json
```diff
@@ -8,8 +8,10 @@
   },
   "scripts": {
     "start": "bundle exec jekyll serve",
-    "prettier": "npx prettier . --write",
+    "prettier": "prettier . --write",
```
tomayac, line 11: I'd leave `npx` here, so `prettier` doesn't have to be globally installed.

tomayac (approved):
LGTM, thanks for this.

Let's have a second pair of eyes review this, though, just to be on the safe side. @dschuff maybe?

package.json
gsnedders: As long as the dev dependencies are installed, then we don't need `npx` here — see https://docs.npmjs.com/cli/v12/commands/npm-run#:~:text=In%20addition%20to,node_modules/.bin%20prefix.

tomayac: You're right, I overlooked it, but it's right there. Resolving then.

gsnedders:
(Force pushed just to rebase past the conflict.)

features.json
```diff
@@ -21,6 +21,11 @@
       "url": "https://github.com/WebAssembly/bulk-memory-operations/blob/master/proposals/bulk-memory-operations/Overview.md",
       "phase": 5
     },
+    "callTags": {
+      "description": "Call Tags",
+      "url": "https://github.com/WebAssembly/call-tags",
+      "phase": "inactive"
```
dschuff: is having "inactive" proposals in the JSON new? I assume we aren't going to show them in the page?

scripts/update-features.js
```diff
@@ -0,0 +1,303 @@
...
+    reportProblem(
+      'Parsed zero proposals from README.md - the "Phase N" heading format may have changed upstream'
+    );
+  }
+
+  return [
+    ...readmeProposals,
+    ...parseFixedPhaseFile(readProposalsFile('finished-proposals.md'), 5),
+    ...parseFixedPhaseFile(
+      readProposalsFile('inactive-proposals.md'),
```
dschuff: This is sort of pre-existing but I'm not sure it's valuable to have inactive proposals. By definition nobody is working on them, and having information won't be useful to developers. I think when I originally made this page the bar was that we don't show anything that doesn't have an implementation, because this page targets developers who might want to target wasm. The scope has expanded some but I'm still not convinced that there's any value in showing proposals with nobody working on them and/or no implementations.
Deprecated proposals would be the exception there, so maybe we make some kind of carveout for that, but I think all the others would be just noise. (or maybe we leave deprecated proposals out too, just to avoid encouraging anyone to use them).

dschuff (commented):
The commit message and comments/documentations should be clear that the keeping-up-to-date part only applies to proposal phases, not engine support.

As for which proposals to show, see my comment below; I think it's still worth curating this a bit; this page doesn't have the same purpose as the wasm proposals repo itself.

scripts/update-features.js
gsnedders: From my point-of-view, I really just want to be able to find all WASM proposals without having to (potentially in multiple places) parse the Markdown files myself — the webpage itself is actually not really my main concern, the JSON file itself is.

The [current (main) features.json](https://github.com/WebAssembly/website/blob/a31b9452495232f634d95defee2eed3f7a4fcfb8/features.json) seems to include all active proposals, plus inactive-but-implemented ones. The fixup commit I've just pushed filters the two places that render things based on features.json down to the that.

